PR #5, opened 2025-10-02
Set up crates for work

chrysn:


chrysn:
Replaced with #6 which hopefully succeeds in triggering actions (this off-origin PR did not until that was sent)

PR #6, opened 2025-10-02
Set up crates for work

chrysn:


chrysn:
Merging this without hax issues resolved: Those will likely need a bit of time from the hax team to sort out.

PR #8, opened 2025-11-20
Architecture: Align with implementation progress

chrysn:


chrysn:
Now includes comments after @jschneider-bensch found gaps in the documentation.

PR #9, opened 2025-12-01
SHA256 for nRF54L15

WilliamTakeshi:
This PR introduces hardware-accelerated SHA-256 for the nRF54L15.

The implementation still contains some duplication and a few magic numbers because it was translated from the vendor’s C code, and the CRACEN documentation appears to be incomplete.

There is also room for improvement in the “oneshot” API. Doing directly with less copying.

* For testing, I added default-members so that cargo test at the workspace root does not attempt to run the nRF54L15 example. Is this the correct approach? I might need to check how other projects handle chip-specific code.

Let me know if you need explanations about any part of the code. It was really hard to understand and adapt from the C examples.

### DIFF
--- a/embedded-cal-nrf54l15/Cargo.toml
+++ b/embedded-cal-nrf54l15/Cargo.toml
@@ -19,7 +19,7 @@ harness = false
 [dependencies]
 embedded-cal.path = "../embedded-cal"
 testvectors.path = "../testvectors"
-nrf54l15-app-pac = { git = "https://github.com/WilliamTakeshi/nrf54l15-app-pac", version = "0.1.0", features = ["critical-section", "rt"] }
+nrf-pac = { version = "0.2.0", features = ["nrf54l15-app", "rt"] }
 
 [dev-dependencies]
 cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
--- a/embedded-cal-nrf54l15/src/lib.rs
+++ b/embedded-cal-nrf54l15/src/lib.rs
@@ -2,6 +2,7 @@
 mod descriptor;
 
 use descriptor::DescriptorChain;
+use nrf_pac::{cracen, cracencore};
 
 const BLOCK_SIZE: usize = 128;
 const MAX_DESCRIPTOR_CHAIN_LEN: usize = 4;
@@ -14,20 +15,17 @@ const _: () = {
 pub struct Nrf54l15Cal {
     // TODO: No need to enable and take ownership of everything
     // it's possible to have a more granular ownership
-    cracen: nrf54l15_app_pac::GlobalCracenS,
-    cracen_core: nrf54l15_app_pac::GlobalCracencoreS,
+    cracen: cracen::Cracen,
+    cracen_core: cracencore::Cracencore,
 }
 
 impl Nrf54l15Cal {
-    pub fn new(
-        cracen: nrf54l15_app_pac::GlobalCracenS,
-        cracen_core: nrf54l15_app_pac::GlobalCracencoreS,
-    ) -> Self {
+    pub fn new(cracen: cracen::Cracen, cracen_core: cracencore::Cracencore) -> Self {
         // Enable cryptomaster
         cracen.enable().write(|w| {
-            w.cryptomaster().set_bit();
-            w.rng().set_bit();
-            w.pkeikg().set_bit()
+            w.set_cryptomaster(true);
+            w.set_rng(true);
+            w.set_pkeikg(true)
         });
 
         Self {
@@ -41,9 +39,9 @@ impl Drop for Nrf54l15Cal {
     fn drop(&mut self) {
         // Disable cryptomaster on drop
         self.cracen.enable().write(|w| {
-            w.cryptomaster().clear_bit();
-            w.rng().clear_bit();
-            w.pkeikg().clear_bit()
+            w.set_cryptomaster(false);
+            w.set_rng(false);
+            w.set_pkeikg(false)
         });
     }
 }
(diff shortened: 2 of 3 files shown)


embedded-cal-nrf54l15/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dependencies]
+cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
+cortex-m-rt = { version = "0.7", features = ["device"] }
+defmt = "1.0"
+defmt-rtt = "1.0"
+panic-probe = { version = "1.0", features = ["print-defmt"] }
+embedded-cal.path = "../embedded-cal"
+testvectors.path = "../testvectors"
+nrf54l15-app-pac = { git = "https://github.com/WilliamTakeshi/nrf54l15-app-pac", version = "0.1.0", features = ["critical-section", "rt"] }
```
malishav: Still no official PAC available for nRF54L15?

embedded-cal-nrf54l15/memory.x
```diff
@@ -0,0 +1,5 @@
+MEMORY
+{
+  FLASH : ORIGIN = 0x00000000, LENGTH = 1524K
+  RAM : ORIGIN = 0x20000000, LENGTH = 128K
```
malishav: At https://www.nordicsemi.com/Products/nRF54L15, I see 1.5 MB flash, but there seems to be 256 kB of RAM. Why have 128 kB as length here?

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,396 @@
+#![no_main]
```
chrysn: What purpose does that serve in a lib?

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, Clone, Copy, defmt::Format)]
+pub struct Descriptor {
+    pub addr: *mut u8,
+    pub next: *mut Descriptor,
+    pub sz: u32,
+    pub dmatag: u32,
```
chrysn:
```suggestion
    addr: *mut u8,
    next: *mut Descriptor,
    sz: u32,
    dmatag: u32,
```

(Or: Why do these need to be pub?)

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+            Some(_) => [instance.algorithm as u8, 0x04, 0x00, 0x00],
+            None => [instance.algorithm as u8, 0x06, 0x00, 0x00],
+        };
+
+        let mut in_desc = Descriptor::empty();
+        let mut state_desc = Descriptor::empty();
+        let mut data_desc = Descriptor::empty();
+        let mut padding_desc = Descriptor::empty();
+
+        match instance.state {
```
chrysn: That looks familiar from above. Can't the padding data be created, sent into `self.update()`, with an assertion that (by construction of the padding) there is no data left in the buffer after it?

chrysn (commented):
Thanks; some initial comments.

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+}
+
+impl embedded_cal::HashAlgorithm for HashAlgorithm {
+    fn len(&self) -> usize {
+        match self {
+            HashAlgorithm::Sha256 => 32,
+        }
+    }
+
+    // FIXME: See from_cose_number definition
```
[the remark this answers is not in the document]
WilliamTakeshi: @chrysn , on this case, the value shouldn't be [-16 instead of -10](https://www.iana.org/assignments/cose/cose.xhtml#algorithms)?

embedded-cal-nrf54l15/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+[[test]]
+name = "integration"
+harness = false
+
+[dependencies]
+cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
+cortex-m-rt = { version = "0.7", features = ["device"] }
+defmt = "1.0"
+defmt-rtt = "1.0"
+panic-probe = { version = "1.0", features = ["print-defmt"] }
```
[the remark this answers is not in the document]
WilliamTakeshi: Fixed on https://github.com/lake-rs/embedded-cal/pull/9/commits/107489c4944e89b4a6e6dcf4eac1cb3513b69e4c

embedded-cal-nrf54l15/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+test = false
+doc = false
+
+[[test]]
+name = "integration"
+harness = false
+
+[dependencies]
+cortex-m = { version = "0.7", features = ["critical-section-single-core"] }
+cortex-m-rt = { version = "0.7", features = ["device"] }
```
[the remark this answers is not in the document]
WilliamTakeshi: Fixed on https://github.com/lake-rs/embedded-cal/pull/9/commits/107489c4944e89b4a6e6dcf4eac1cb3513b69e4c

embedded-cal-nrf54l15/src/lib.rs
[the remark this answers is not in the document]
chrysn: Yes indeed; there's a comment in `embedded-cal/src/lib.rs` even about this. My use of -10 stems from me misremembering the trouble about algorithms -- trouble is with specifying *a HKDF algorithm*, whereas hash algorthms are fine.

See https://github.com/lake-rs/embedded-cal/pull/14

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,396 @@
+#![no_main]
+#![no_std]
+
+use defmt_rtt as _;
+use panic_probe as _;
+
+const BLOCK_SIZE: usize = 64;
```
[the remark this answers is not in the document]
WilliamTakeshi: updated to work with SHA384 and SHA512, yes, the BLOCK_SIZE now is 128 and I added some compile time checks

embedded-cal-nrf54l15/src/lib.rs
chrysn: The issue (that AIU still persists) is that there is a large amount of duplication between the update and the finish function. This would be way more readable and less prone to editing errors if some of that update/finish code were refactored into (non-trait) methods that are then used commonly between update and finalize.

embedded-cal-nrf54l15/src/lib.rs
chrysn: Why do they even need to be pub(crate)? So far, all code that is accessing them is within the same module.

embedded-cal-nrf54l15/src/lib.rs
```diff
@@ -0,0 +1,358 @@
...
+            HashAlgorithm::Sha384 => 48,
+            HashAlgorithm::Sha512 => 64,
+        }
+    }
+
+    // FIXME: I am using the values from RFC 9054 here, is that correct?
+    // https://www.rfc-editor.org/rfc/rfc9054.html#name-sha-2-hash-algorithms
+    fn from_cose_number(number: impl Into<i128>) -> Option<Self> {
+        match number.into() {
+            -10 => Some(HashAlgorithm::Sha256),
```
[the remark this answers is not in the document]
chrysn: That was my mistake from mixing up SHA256 with HKDF based on SHA256; this was already discussed (and presumably fixed?) in https://github.com/lake-rs/embedded-cal/pull/9#discussion_r2618814755.

Issue #12, opened 2025-12-10
Update architecture: SHA256 also has not-a-full-algorithm building blocks


chrysn:
See <https://github.com/lake-rs/embedded-cal/pull/9/files#r2592974180> for whence this comes from.

We did plan to use AES as a building block for CCM, but not block-only hashes that leave it to an upper layer to do spooling (which can be algorithm independent) and filling the last bytes (which is algorithm dependent).

* [ ] Describe API
* [ ] Provide implementation
* [ ] Update #9 (likely merged by then) to use it

PR #16, opened 2025-12-15
hash: Add block-only algorithm constructor

chrysn:
Closes: https://github.com/lake-rs/embedded-cal/issues/12

There's some way to go to provide the top layer, but for a first check, @WilliamTakeshi: If you'd only provide the limited version introduced here for https://github.com/lake-rs/embedded-cal/pull/9, you could just drop all the buffers, and wouldn't need to deal with padding at finalization, right?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -76,6 +76,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "embedded-cal-software"
+version = "0.1.0"
+dependencies = [
+ "embedded-cal",
+ "testvectors",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["embedded-cal", "embedded-cal-rustcrypto", "testvectors"]
+members = ["embedded-cal", "embedded-cal-rustcrypto", "embedded-cal-software", "testvectors"]
 resolver = "3"
 
 [workspace.package]
--- a/embedded-cal-software/Cargo.toml
+++ b/embedded-cal-software/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "embedded-cal-software"
+edition.workspace = true
+license.workspace = true
+readme.workspace = true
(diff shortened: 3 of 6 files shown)


chrysn:
With the 2nd commit I've pulled the padding out of #9. I wouldn't worry about how this is done prettily, as all that code in embedded-cal-software would eventually be replaced with libcrux, but it'll allow you to test doing #9 without any of the buffer and padding management.

chrysn:
@jschneider-bensch, would this interface work libcrux?

And in particular: Could libcrux implement this in a way that even when using libcrux as a full SHA2 implementation, it could still provide the "bare SHA" (only full blocks, no padding) and "wrapper around bare" parts separately, *and still let the proofs run through*? I'm not yet sure how critical that is, but I'd rather check now and adjust if needed. If nothing else, splitting also the libcrux version along this boarder will make it easier to opt in to hardware acceleration where it is supported with less configuration effort.

jschneider-bensch:
In `libcrux-sha2`, with feature `expose-hacl` we expose the low level functions that are otherwise wrapped by the high-level `init/update/finish` API, and we could conceivably implement a full-block oriented hash using these.

Regarding proofs: When we take an implementation exported from HACL (or HACL-rs), as in the case of SHA-2, we don't have a prover in the loop, so to say. The idea here is instead that HACL* is a verified implementation in F*, which can be exported to C, or Rust preserving the verification results established on the F* side, **if** the pre-conditions for calling any of the functions established there are maintained. So in libcrux, we leave the exported code untouched and give wrappers that ensure the pre-conditions are met and which implement the high-level API. If the low-level functions are called directly, it has to be ensured that all preconditions are met.

I have a basic question here, though: Reading through #9, I understood that the motivation for adding a block-only version is too get rid of the padding and spooling code in the provider, right? Would this then be done generically, independent of the backend?

WilliamTakeshi:
I did some quick tests here:
if we use this limited API, with only full blocks, I **THINK** I can create a implementation where we can drop the `block` and `block_size` from  the `HashState`, but I cannot drop the padding.
Even for full-blocks, the SHA needs the padding to work.

Maybe I am missing something, but I still didn't understand the reason of this more constrained API. Is the `block` and the `padding` bad? They break something on the formal verification?

PR #29, opened 2026-03-05
chore: add from_ni_name to embedded-cal-software

WilliamTakeshi:
Trying to use `embedded-cal` I discovered that `from_ni_name` is not implemented.


chrysn (approved):
Thanks for catching this.

I briefly considered whether, as an alternative, we should have provided functions `from_ni_{name,number}` that map them all to the COSE number, but seeing that not all NI hashes are also recognized in COSE (eg. SHA-256/32, and also blake?), so there's not one we can easily provide.

Issue #26, opened 2026-03-03
[tracking] Possible enhancements after nRF54 merge


chrysn:
There are a few things I think might have made sense in #19/#25, but we did leave them for after merging.

Please take them as things to consider / suggestions, not as necessary-to-implement items:

* There are a lot of magic constants. Pull them out, give them tentative names, and document somewhere there that they are a product of reverse engineering.
    * In particular, there are more occurrences of the 0x2000_0000 constant, which might be refactored into the sz() function.
* If there is no aliasing allowed between input and output descriptors, make the descriptors take a slice rather than a pointer, use sz() as a private-only function inside the descriptor handling.

  This probably requires type-level distinction between input and output buffers, as the latter take a &mut slice.

  It will mean that the borrow checker can see that indeed that memory is used until that descriptor is consumed.

  As things are currently, the descriptors contain raw pointers (which don't need unsafe code to generate), but then write there, which means that there should be unsafe declarations somewhere (eg. around `execute_cryptomaster_dma`); most of them are to have a good place to describe the relevant invariants. With the refactoring above, those would be more internal and short-lived.
* If there *is* aliasing allowed (eg. we use the same memory to pass in the inital state and the output state), descriptor chains could have a pair-of-descriptor-chains type whose life time captures both parts, and that has a `.push_input_and_mut_output(slice: &mut [u8])` that internally gets turned into two pointers. But even if it is allowed, as long as we don't need it, we don't need it implemented that way (and having the state on the stack twice is something I'd be OK with if it means that the code stays simpler).
* A descriptor chain is self-referential through the next field of the descriptors. Ways to fix this:
  - Don't set .next() at push time, but before using, eg. in `.first()`
  - Use it pinned.
  
  Relatedly, DescriptorChain::first is a bit brittle in that it does not guarantee anything about the lifetime of the pointer/register-value. Things work in execute_cryptomaster_dma because that takes a &mut DescriptorChain which guarantees that those will be live until the end of the function, but if that were carelessly refactored into taking the DescriptorChain by value, it's up to the optimizer whether or not the DescriptorChain was dropped. I haven't thought through all options of how to do that best, but one is to use it as `.with_first_pointer(|ptr: u32| { dma…write_value((ptr); dma…wait for completion; })`.

PR #30, opened 2026-03-06
nRF54 typed descriptor chains

WilliamTakeshi:
enhancements for the nRF54 code. https://github.com/lake-rs/embedded-cal/issues/26

This PR is divided in 3 commits:
* First one, create diff types for Input and Output in `DescriptorChain`s, it should make it type safer for creating and using this API
* Updates all the `.next` field just before calling `first()`. It should make it safer to protect against moves
* Add `with_first_pointer` to hold the `DescriptorChain` pinned in memory while is being used.

### DIFF
--- a/embedded-cal-nrf54l15/src/descriptor.rs
+++ b/embedded-cal-nrf54l15/src/descriptor.rs
@@ -121,6 +121,15 @@ impl<Direction, const N: usize> DescriptorChain<Direction, N> {
         self.update_links();
         &self.descs[0] as *const Descriptor as u32
     }
+
+    /// Calls `f` with the address of the first descriptor, holding a `&mut self`
+    /// borrow for the duration of the call.
+    ///
+    /// This guarantees that the descriptor chain remains live and pinned in memory
+    /// while `f` is executing, so the pointer passed to hardware stays valid.
+    pub(crate) fn with_first_pointer(&mut self, f: impl FnOnce(u32) -> ()) {
+        f(self.first())
+    }
 }
 
 impl<const N: usize> DescriptorChain<Input, N> {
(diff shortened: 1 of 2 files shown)


embedded-cal-nrf54l15/src/descriptor.rs
```diff
@@ -71,58 +81,94 @@ impl<const N: usize> DescriptorChain<N> {
...
+impl<const N: usize> DescriptorChain<Input, N> {
+    /// Appends an input (read) buffer to the chain.
+    ///
+    /// The DMA engine will read from `data` during the transfer.
+    ///
+    /// # Safety / Correctness requirements
+    ///
+    /// - `data` must be DMA-accessible memory.
+    /// - The chain must not be mutated after being handed to the EasyDMA
+    ///   hardware until the END or ERROR event is observed.
```
chrysn: Rather than making this a requirement that is a bit intransparent to the user, can we encode this in the type? For now that a slice is passed in, the expectation would be justified that its lifetime is accurate.

Like, would that work?: Make it a `DescriptorChain<'mem, Input, N>` that takes a `push(&mut self, data: &'mem [u8], …)`, and that 'mem can be arbitrarily short (ie. a user can drop an incomplete Descriptor), but once it is passed to a DMA operation, that operation will take the 'mem lifetime and uphold the END/ERROR constraints (eg. using drop guards).

The requirement could then be in `with_first_pointer`, namely that `f must only return once END or ERROR was seen, and can only panic before the data is sent to a DMA engine or after END / ERROR was seen`. (It's not really a SAFETY requirement *there*, the requirement is in `dma.{fetch,push}addrlsb().write_value(…);`, but as this module's purpose is to make safety requirements actionable, this is a good place).

chrysn (commented):
Generally an improvement, but I have suggestions around lifetimes which I'm trying out right now…

chrysn (approved):
The commit I pushed would address my concerns. Marking the PR as ACKed, but waiting for your OK for the added commit before merging.

chrysn:
Thanks for checking variance, I forgot to do that thoroughly.

WilliamTakeshi:
Looks good! I just made one change. I think we need to add `*mut` before to make it invariant.

Because my understanding is:
'mem of data can be shorter when DescriptorChain is used as Input but must be longer on DescriptorChain is used as Output

PR #32, opened 2026-03-13
fix bug buffer len when FIRST_CHUNK_SIZE > SHA2SHORT_BLOCK_SIZE

WilliamTakeshi:
Fix incorrect `written_in_buffer` calculation in both `update()` and `finalize()` when a backend has `FIRST_CHUNK_SIZE > SHA2SHORT_BLOCK_SIZE` (e.g. STM32WBA55: 68 > 64)

When buffer is between [SHA2SHORT_BLOCK_SIZE, FIRST_CHUNK_SIZE) the old code didn't calculated correctly `written_in_buffer` when doing `%= SHA2SHORT_BLOCK_SIZE`. For example when finalizing with 64 bytes, it send `[]` as the `last_chunk`


chrysn (approved):
Thanks, good catch.

I hope we don't have to spend too much time with this code anyway with the incoming libcrux based proper version.

PR #33, opened 2026-04-03
feat: Include rand_core::TryCryptoRng in traits

WilliamTakeshi:
This PR add the `TryRng` trait and hardware impl for nRF54L15 and STM32WBA55

The `TryRng` was based on the `rand_core::TryRng` trait (should we just add as a dependency and impl?)

Summary

  - Introduces a TryRng trait in embedded-cal (modelled after rand_core::TryRng) with try_next_u32, try_next_u64, and try_fill_bytes methods, plus a shared test_tryrng helper for back-end
  validation
  - Implements TryRng for the nRF54L15 back-end (embedded-cal-nrf54l15) using the CRACEN RNG hardware
  - Implements TryRng for the STM32WBA55 back-end (embedded-cal-stm32wba55) using the on-chip RNG peripheral

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,6 +197,7 @@ name = "embedded-cal"
 version = "0.1.0"
 dependencies = [
  "hax-lib",
+ "rand_core",
 ]
 
 [[package]]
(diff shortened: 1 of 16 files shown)


chrysn:
Thanks. I think that given how usable the rand_core::TryRng is, we can just add the foreign trait to our trait dependencies. (Actually rand_core::TryCryptoRng, because a non-CS RNG is not really within our scope).

PR #34, opened 2026-04-07
feat: clippy on all workspaces

WilliamTakeshi:
After splitting the workspace into `members` and `default-members` to not run tests on every workspace, it also made clippy only run on default members. This PR fix that.

clippy::needless-range-loop, clippy::into-iter-on-ref, unused_variables, unnecessary-cast, unused-unit, manual-is-multiple-of

### DIFF
--- a/embedded-cal-stm32wba55/src/lib.rs
+++ b/embedded-cal-stm32wba55/src/lib.rs
@@ -155,7 +155,7 @@ impl Stm32wba55Cal {
 
         // Save CSR registers (0..37 always; 38..53 only for HMAC)
         let mut csr = [0u32; CSR_REGS_LEN];
-        for (i, slot) in csr.iter_mut().enumerate().take(CSR_REGS_LEN) {
+        for (i, slot) in csr.iter_mut().enumerate() {
             *slot = self.hash.csr(i).read();
         }
         instance.context = Some(Context { csr, str, imr });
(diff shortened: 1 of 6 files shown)


embedded-cal-stm32wba55/src/lib.rs
```diff
@@ -156,8 +156,8 @@ impl Stm32wba55Cal {
 
         // Save CSR registers (0..37 always; 38..53 only for HMAC)
         let mut csr = [0u32; CSR_REGS_LEN];
-        for i in 0..CSR_REGS_LEN {
-            csr[i] = self.hash.csr(i).read();
+        for (i, slot) in csr.iter_mut().enumerate().take(CSR_REGS_LEN) {
```
chrysn: Why take() that many? The iterator is already that length by construction.

```suggestion
        for (i, slot) in csr.iter_mut().enumerate() {
```

embedded-cal-stm32wba55/src/lib.rs
```diff
@@ -205,8 +205,8 @@ impl Stm32wba55Cal {
     }
 
     fn read_digest(&mut self, out: &mut [u32; 8]) {
-        for i in 0..8 {
-            out[i] = self.hash.hr(i).read();
+        for (i, slot) in out.iter_mut().enumerate().take(8) {
```
chrysn: As above.
```suggestion
        for (i, slot) in out.iter_mut().enumerate() {
```

chrysn (commented):
Thanks, LGTM, with two cases of "Was this a Clippy auto-suggestion where it didn't take all context into account?"

WilliamTakeshi:
Rebased and remove unnecessary .take()

PR #37, opened 2026-04-23
initial libcrux sha256

franziskuskiefer:
Here's a first version for using libcrux sha256.
I'm not entirely sure if I understood everything right, but the tests seem to pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -196,7 +196,7 @@ dependencies = [
 name = "embedded-cal"
 version = "0.1.0"
 dependencies = [
- "hax-lib",
+ "hax-lib 0.3.5",
 ]
 
 [[package]]
@@ -230,6 +230,7 @@ name = "embedded-cal-software"
 version = "0.1.0"
 dependencies = [
  "embedded-cal",
+ "libcrux-sha2",
  "testvectors",
 ]
 
@@ -286,7 +287,18 @@ name = "hax-lib"
 version = "0.3.5"
 source = "git+https://github.com/hacspec/hax#bf9a7e4b6a24aebf015910d62477103d010fd643"
 dependencies = [
- "hax-lib-macros",
+ "hax-lib-macros 0.3.5",
+ "num-bigint",
+ "num-traits",
+]
+
+[[package]]
+name = "hax-lib"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "543f93241d32b3f00569201bfce9d7a93c92c6421b23c77864ac929dc947b9fc"
+dependencies = [
+ "hax-lib-macros 0.3.6",
  "num-bigint",
  "num-traits",
 ]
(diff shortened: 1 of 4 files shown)


WilliamTakeshi:
Hey, thanks for the PR.

I don't have access to the boards right now, so I cannot confirm everything, but I think this PR breaks sha256 for both boards.

The code inside `embedded-cal-software/src/lib.rs` is used for both nrf54 and stm32.

you should implement like `embedded-cal-software/src/tests/dummy_sha256.rs` and make the `Extender` just be a thin wrapper on top of it.

If you have a board you can `cargo test` inside `embedded-cal-nrf54l15/` or `embedded-cal-stm32wba55/`

franziskuskiefer:
I don't have access to boards. Let me know if you know what breaks and why.

chrysn:
The bigger issue than concrete breakage is where this is between being a software-only implementation and using hardware accelerators.

This sets out to replace -software (the dummy stand-in for libcrux) and not -rustcrypto. The difference between them is that the former has so far (and should -- that's part of the premise of this project) pick up the lower-level operations that hardware accelerators have, whereas the latter can be a reference software-only version. (-software can also be a software-only version when built not on the dummy which mocks a crypto accelerator, but built on a `T` where `<T as Sha2Short>::SUPPORTED == false`).

dummy_sha256 should not need to be touched for this; it is just a testing helper that allows us to switch around between having or not having hardware acceleration without actually pulling in a hardware accelerator.

The libcrux implementation should use the primitives provided through any implementation of Sha2Short, whereas currently it does nothing in that direction.

franziskuskiefer:
> The bigger issue than concrete breakage is where this is between being a software-only implementation and using hardware accelerators.

I don't think this is what we discussed or planned for. The libcrux implementation won't call into hardware acceleration. It's a software only implementation. If you want to do something else, you'll need to write and verify new code here.

WilliamTakeshi:
@franziskuskiefer 

The `Extender` is just a thin brigde to gap the difference between hardware backends (like align-blocks, first block size, padding, etc.). The idea is that on the case of the software implementation, will change nothing, it will be a straight call to `Sha2Short::init`, `Sha2Short::update`, and `Sha2Short::finalize`.

So I think the right path here is having the libcrux implement like the dummy, just implement Sha2Short directly and the `Extender` will wrap without adding anything

chrysn:
<del>I did not check the original project documents for commitment, but we discussed this when the plumbing design was changed to accommodate your recommendations. If libcrux does not tap into hardware acceleration through embedded-cal, the usefulness of the combination of embedded-cal and libcrux deteriorates from allowing the use of formally verified software to accelerate any vendor's hardware, to using libcrux next to them, like one might also use rust-crypto implementations -- still useful, but by far not what it could be.</del>

[edit, after personal chat]: We do plan to use libcrux with hardware acceleration, but the initial operations we do this for will be ECDH (and/or ECDSA), where it pays off more; we may still add that later.

PR #38, opened 2026-05-19
Add HmacProvider trait and implementation for SHA256

WilliamTakeshi:
* add `HmacProvider` and `HmacAlgorithm` traits, mirroring the `HashProvider`/`HashAlgorithm` that we already have
* impl a software HMAC-SHA-256 in the `Extender` built on top of the `HashProvider`
* add some test cases for the [RFC 4231](https://datatracker.ietf.org/doc/html/rfc4231#section-4.2)
* impl the hardware accelerated hmac-sha-256 for stm32

### DIFF
--- a/embedded-cal-software/src/lib.rs
+++ b/embedded-cal-software/src/lib.rs
@@ -5,7 +5,7 @@
 #![no_std]
 
 use embedded_cal::{
-    Cal, HashProvider, HmacProvider,
+    HashProvider, HmacProvider,
     plumbing::Plumbing,
     plumbing::hash::{SHA2SHORT_BLOCK_SIZE, Sha2Short, Sha2ShortVariant},
 };
@@ -13,7 +13,7 @@ use embedded_cal::{
 pub trait ExtenderConfig {
     const IMPLEMENT_SHA2SHORT: bool;
 
-    type Base: Cal + Plumbing;
+    type Base: HashProvider + Plumbing;
 }
 
 impl<EC: ExtenderConfig> Extender<EC> {
--- a/embedded-cal-software/src/lib.rs
+++ b/embedded-cal-software/src/lib.rs
@@ -422,4 +422,11 @@ mod tests {
 
         testvectors::test_hash_algorithm_sha256(&mut cal);
     }
+
+    #[test]
+    fn test_hmac_sha256_on_dummy() {
+        let mut cal = Extender::<ImplementSha256Short>(dummy_sha256::DummySha256);
+
+        testvectors::test_hmac_sha256(&mut cal);
+    }
 }
--- a/embedded-cal-software/src/lib.rs
+++ b/embedded-cal-software/src/lib.rs
@@ -5,7 +5,7 @@
 #![no_std]
 
 use embedded_cal::{
-    Cal, HashProvider,
+    Cal, HashProvider, HmacProvider,
     plumbing::Plumbing,
     plumbing::hash::{SHA2SHORT_BLOCK_SIZE, Sha2Short, Sha2ShortVariant},
 };
@@ -123,7 +123,7 @@ impl<EC: ExtenderConfig> HashProvider for Extender<EC> {
                         buffer,
                         instance,
                     };
-                    self.update(&mut rewrapped, &padding[..padding_size]);
+                    HashProvider::update(self, &mut rewrapped, &padding[..padding_size]);
                     let HashState::Sha256 {
                         instance, buffer, ..
                     } = rewrapped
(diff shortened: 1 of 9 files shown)


embedded-cal/src/hmac.rs
```diff
@@ -0,0 +1,71 @@
+pub trait HmacProvider {
+    type Algorithm: HmacAlgorithm;
+    /// State carried between rounds of feeding data into the HMAC.
+    type HmacState: Sized;
+    /// Output of an HMAC operation.
+    type HmacResult: AsRef<[u8]>;
```
chrysn, line 6: We should have one more in here: `Key`, for not all HMAC operations should start from the key being visible as a `&[u8]` to the user. They key might be a random output (eg. when a system hands off data to an outsider to later consume in a trusted way, like CoAP extended tokens), or it might be the output of a KDF. Where hardware supports it, that key material might not even ever be in memory.

That key needs to be clonable (whereas the HmacState generally is not), so we can't just equate those two (even though the init function could be used that way).

No need to fix this in this PR, I'd do a follow-up, as a `.init()` function from keys in RAM is a sensible shortcut.

embedded-cal-software/src/tests/dummy_sha256.rs
```diff
@@ -42,6 +42,24 @@ impl embedded_cal::HashProvider for DummySha256 {
     }
 }
 
+impl embedded_cal::HmacProvider for DummySha256 {
```
chrysn, line 43: These are the 20 lines mentioned in https://github.com/lake-rs/embedded-cal/issues/40 -- but as said there: not now.

embedded-cal-software/src/lib.rs
```diff
@@ -123,7 +123,7 @@ impl<EC: ExtenderConfig> HashProvider for Extender<EC> {
                         buffer,
                         instance,
                     };
-                    self.update(&mut rewrapped, &padding[..padding_size]);
+                    HashProvider::update(self, &mut rewrapped, &padding[..padding_size]);
```
chrysn, line 152: … and that is the usage ugliness from https://github.com/lake-rs/embedded-cal/issues/40 (also, good for now)

embedded-cal-software/src/lib.rs
```diff
@@ -259,6 +259,115 @@ impl<EC: ExtenderConfig> AsRef<[u8]> for HashResult<EC> {
     }
 }
 
+/// HMAC algorithm identifier for software HMAC over [`Extender`].
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum HmacAlgorithm {
+    HmacSha256,
```
chrysn, line 291: As with HashResult, this could extend the underlying type's algorithms. Unlike there, we don't have any hardware that currently provides it, but on the long run this will matter because if the underlying library encapsulates secrets, we'd like it to stay in there.

But this can be extended in a follow-up.

embedded-cal-software/src/lib.rs
```diff
@@ -259,6 +259,115 @@ impl<EC: ExtenderConfig> AsRef<[u8]> for HashResult<EC> {
...
+    fn from_cose_number(number: impl Into<i128>) -> Option<Self> {
+        match number.into() {
+            5 => Some(HmacAlgorithm::HmacSha256),
+            _ => None,
+        }
+    }
+}
+
+pub enum HmacState<EC: ExtenderConfig> {
+    HmacSha256 {
```
chrysn, line 311: Can we generalize this to not carry one of those for every algorithm? The only hash algorithm dependent part here is the outer_key length, and that can be a maximum (and we carry a copy of the algorithm around from which the size can be inferred -- or we allow the algorithm to be reconstructed from the state, which might be a good feature in general).

Can be in a follow-up.

embedded-cal-stm32wba55/src/lib.rs
```diff
@@ -57,21 +61,176 @@ impl embedded_cal::HashProvider for Stm32wba55Cal {
...
+        match algorithm {
+            HmacAlgorithm::HmacSha256 => {
+                // Normalise key: zero-pad short keys; hash long keys per RFC 2104.
+                let mut key_block = [0u8; SHA2SHORT_BLOCK_SIZE];
+                if key.len() > SHA2SHORT_BLOCK_SIZE {
+                    let hashed = self.sha256_of(key);
+                    key_block[..SHA256_OUT_BYTES].copy_from_slice(&hashed);
+                } else {
+                    key_block[..key.len()].copy_from_slice(key);
+                }
```
chrysn, line 210: This looks like something we might pull out, but it's too small to matter a lot or warrant a plumbing layer, or even a "init with suitably sized key" method where an "init with any key" is provided. (So I'd say no action now).

embedded-cal-stm32wba55/src/lib.rs
```diff
@@ -57,21 +61,176 @@ impl embedded_cal::HashProvider for Stm32wba55Cal {
...
+}
 
-    fn finalize(&mut self, state: Self::HmacState) -> Self::HmacResult {
-        match state {}
+/// State for an in-progress HMAC-SHA256 operation.
+pub struct HmacState {
+    /// Saved HASH hardware context (captured after each block written to hardware).
+    context: Option<Context>,
+    /// Buffer for accumulating a full 64-byte (68 on the first call) block before sending to hardware.
+    buf: [u8; HMAC_MAX_BLOCK_SIZE],
```
chrysn, line 178: Hmpf, this looks like it would warrant having some layering like with Hash that allows the buffer handling to be done by generic code. But that'll need more thought, and this works now. (I'll create a "things to be followed up on" issue).

embedded-cal-stm32wba55/src/lib.rs
```diff
@@ -204,6 +369,56 @@ impl Stm32wba55Cal {
         }
     }
 
+    /// Resume an interrupted HMAC-SHA256 operation.
+    ///
+    /// Identical to [`restore_context`] but sets `MODE = 1` (HMAC) and `LKEY = 0`.
+    fn restore_context_hmac(&mut self, ctx: &Context) {
```
chrysn, line 463: Why not `fn restore_context_hmac(&mut self, ctx: &Context) { self.restore_context(ctx, 1, 0) }` or `fn restore_context<MODE: bool, LKEY: bool>(&mut self: ctx: &Context<MODE, LKEY>)` ?

embedded-cal-stm32wba55/tests/integration.rs
```diff
@@ -11,7 +11,10 @@ impl embedded_cal_software::ExtenderConfig for ImplementSha256Short {
 }
 
 struct TestState {
+    /// Software-extended CAL used for the SHA-256 hash tests.
     cal: embedded_cal_software::Extender<ImplementSha256Short>,
+    /// Raw STM32WBA55 CAL used to exercise the hardware HMAC accelerator directly.
+    raw: embedded_cal_stm32wba55::Stm32wba55Cal,
```
chrysn, line 17: I think we could simplify here if we just let the software CAL act as an extender on HMACs. That'd make testing things in lopsided ways (eg. software HMAC on STM32 hardware hash) harder, but a) I don't really see why we should need to (as those are reasonably isolated), and b) if we still want to do that eg. for performance/footprint testing, I'd add a "whiteout" wrapper that allows hiding some algorithms which a type might provide. (Thus, we could test `software::Extender<WhiteoutHkdf<5, Stm32wba55Cal>>`, but it'd be the odd one out.)

chrysn (approved):
LGTM, thanks for adding this.

There are a few things I think we should do better, but let's discuss them in a follow-up issue/PR.

As another PR builds on this, I'll look into what it needs to merge this quickly.

PR #39, opened 2026-05-21
API design improvement

WilliamTakeshi:
I think the PR is ready to review, but since its on top [hmac](https://github.com/lake-rs/embedded-cal/pull/38)

  Summary
- Add an examples folder, this is more for testing how the user will use the API and make sure we don't break anything, plus have examples for the future
- Rename the core crate to embedded-cal-core, and re-export everything throught a embedded-cal crate thin umbrella
- Create a newtype `CalInner` that is the old `Cal`, and the new `Cal` already return everything with the `Extender` so the user doesn't have to do that
  - a bit of moving files around, so create a inner.rs that has the inner, so everything doesn't live on lib.rs. More work should be done in this part still


WilliamTakeshi:
closed, I will work on a new PR later

PR #41, opened 2026-05-25
Test PR, please ignore

chrysn:
Just running CI on a commit which I'd like to push onto https://github.com/lake-rs/embedded-cal/pull/38 before merging, and I think that while GitHub will let me push to that branch, it won't let me force push fixes.


chrysn:
This has served its purpose, now pushed onto #38.

Issue #42, opened 2026-05-25
[tracking] Possible improvements around HMAC


chrysn:
After merging https://github.com/lake-rs/embedded-cal/pull/38, there are some things I'd like to change still (currently trying out some of them):

* [ ] Have type for Algorithm+Key base (to allow constructing an HMAC without ever making the visible, or even present in RAM); see https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297434434

  Proposal in https://github.com/lake-rs/embedded-cal/pull/44.

* [ ] Make the software implementation use the underlying HMAC (see https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297566242 and https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297748540)

* [ ] Software HMAC doesn't need to have an enum state variant for each algorihtm; see https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297605255

* [ ] There are steps that might be shared between hardware implementation that might better be done in a generic way:

  * Leave "what if the key is not perfectly sized" steps to a general component (https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297637041). This might even serve as an optimization for size: If all code paths for HMAC go through an "and we know that the size is right" path (eg. because the keys all come from other exporters that just pick the right length, or from the RNG), that "pad with zeros or hash once" step could be removed by dead-code elimination.

  * There is some amount of buffering (https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297669344)

* [ ] Storing the context might be done differently, see https://github.com/lake-rs/embedded-cal/pull/38#discussion_r3297686190 (less duplication; better store-load mixup safety)

chrysn:
There's one more I found when cleaning up around AEAD:

* [ ] If I'm not mistaken, the software implementation of HMAC goes into the software implementation of Hash with software implementation algorithms, rather than going into Self with Self algorithms.

  That would defeat hardware acceleration, and would also become visible once we allow stripping down the software extender to not implement variants that the hardware provides anyway. (Currently not done for simplicity, but that'd be dead code that the compiler can't remove because it doesn't recognize which enum variants can never be constructed).

PR #43, opened 2026-05-25
Hashes: Require Clone on states

chrysn:
This is no absolute requirement for implementing key items for HMAC (item 1 of https://github.com/lake-rs/embedded-cal/issues/42), but simplifies it.

The main review question is whether that is an operation we *want* to expect. (People could still expect it *locally*, eg. by requiring that the associated type be Clone). It is typically provided, but is that common enough to be universal?

### DIFF
--- a/embedded-cal-nrf54l15/src/lib.rs
+++ b/embedded-cal-nrf54l15/src/lib.rs
@@ -59,6 +59,7 @@ impl Drop for Nrf54l15Cal {
     }
 }
 
+#[derive(Clone)]
 pub struct HashState {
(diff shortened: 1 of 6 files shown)


chrysn:
If the outcome is "we don't want it', I can split up this PR so that at least the software impls do provide it (because they can).

WilliamTakeshi (approved):

chrysn:
Is that just technical ACK or do you also think that Clone is sensible for hash intermediate states throughout?

WilliamTakeshi:
> Is that just technical ACK or do you also think that Clone is sensible for hash intermediate states throughout?

I don't have experience with many different chips, but I think as a user I would expect it to be clonable, so I think its reasonable change

PR #44, opened 2026-05-25
HMAC: Introduce key type

chrysn:
This follows up on item 1 of https://github.com/lake-rs/embedded-cal/issues/42: have a dedicated type, so that encapsulated keys can be used.

It builds on https://github.com/lake-rs/embedded-cal/pull/43, but could be done differently too (it was just easier this way).

It is not completely without alternative: We *could* also decide that any key material that gets taken from elsewhere becomes a HmacState that happens to be otherwise un-updated -- but I think the intention of the APIs is clearer with a dedicated type, the hardware impls show that it is easy to do, and it also allows for some more optimizations (eg. because keys might be kept around in memory between operations, and thus don't need to carry a buffer, even though the current impls happen to do that).

### DIFF
--- a/embedded-cal-nrf54l15/src/lib.rs
+++ b/embedded-cal-nrf54l15/src/lib.rs
@@ -99,11 +99,16 @@ impl embedded_cal::HashProvider for Nrf54l15Cal {
 
 impl embedded_cal::HmacProvider for Nrf54l15Cal {
     type Algorithm = embedded_cal::NoHmacAlgorithms;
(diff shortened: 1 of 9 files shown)


chrysn:
TBD: It probably makes more sense to take the Key by reference, and then we could still think about whether it needs to be Clone or not.

(But postponing making up my mind as I'm on a different task now).

WilliamTakeshi (approved):

PR #45, opened 2026-05-25
Add AEAD interface

chrysn:


WilliamTakeshi (approved):
Looks good to me.

Just a small thing, we should have a note on `decrypt_in_place` to make sure we zero everything in case of errors. I was reading the [rustcrypto decrypt_in_place](https://github.com/RustCrypto/AEADs/security/advisories/GHSA-423w-p2w9-r7vq) and saw that they had problems with this before

chrysn:
Yes, that's a good addition I'll add right in here.

chrysn:
GitHub had some undebuggable hickup where it checked something from main rather than checking this branch; no-op'ly amended the last commit to run CI again.

PR #47, opened 2026-05-27
Clean up after missing CI checks

chrysn:
Something went wrong merging https://github.com/lake-rs/embedded-cal/pull/44 and a formatting error [edit: and a clippy warning and more] slipped through -- more precisely, that's because no branch protection rule enforces that the tests pass, and the visual checks tend to get ignored while F* export is still not working.


chrysn:
Those checks are now requiring; taking the liberty to merge this without further review.

PR #49, opened 2026-05-27
Enable AEAD in Cal

chrysn:
In https://github.com/lake-rs/embedded-cal/pull/45 I missed making AeadProvider a mandatory Cal interface, and thus a bit of fall-out.

As this would have pushed the software implementation parts over the limit of being manageable, it is blocked by #48 (on which it builds).


WilliamTakeshi (approved):
LGTM! 🚀

PR #50, opened 2026-05-27
Add libcrux imlementation of sha256

chrysn:
This is a bis of #37. Initially, it renames embedded-cal-software to -software-demo and adds a README to clean up the misunderstanding due to which the original PR was a change in -software rather than a new crate.

WIP: So far, this is just the renaming (but PRing already for visibility and to get CI checks as we go).


chrysn:
This is now ready for review.

@franziskuskiefer, I hope the new README of the -software (now -software-demo) clarifies the intentions of that. In a sense, it's the staging area where we try out, with a stupid implementation almost copy-pasted from Wikipedia, whether the interfaces work well enough (currently: for SHA2) even if we don't use that exact plumbing part through libcrux.

(Once we go beyond PoC and consider deploying anything where the hardware based hashers come in, we can then still consider whether we can do the libcrux porting, or whether parts of the software-demo code are sufficiently not-sensitive; after all, they do mostly what outside of embedded-cal the author of the software wrapping the hardware implementation would write themselves.)

chrysn:
There's one bug in there that is that the empty impls of this extender should really forward everything rather than being null impls; can fix that later (but am playing a bit with libcrux aes-gcm, and that makes that kind'a moot anyway).

embedded-cal-libcrux/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+license.workspace = true
+readme.workspace = true
+repository.workspace = true
+keywords.workspace = true
+categories.workspace = true
+version.workspace = true
+
+[dependencies]
+embedded-cal.path = "../embedded-cal"
+libcrux-sha2 = "0.0.6"
```
franziskuskiefer:
```suggestion
libcrux-sha2 = "0.0.7"
```

franziskuskiefer (approved):
lgtm with a bump to the latest release

chrysn:
Thanks. Applied with an extra Cargo.lock refresh, squashed the refresh, and merging.

Issue #1, opened 2025-07-01
Initial feature wishlist


chrysn:
In this project, we will build an abstraction layer (as a Rust trait or family of traits) that provides the best of formally verified implementations and hardware cryptography to Rust programs. The actual operations will be performed through established Rust crypto traits; this is about representing a platform's provided operations. (Comparing to embedded-hal and Ariel OS, the existing traits are what embedded-hal is, whereas this project abstracts over what stm32-hal and nrf52-hal do for setting up GPIO pins etc).

In this issue, we collect wishlists of what this should do – well aware that part of all this will also be setting scope, but that doesn't need to stop us from wishes coming in.

My personal baseline, informed from the project setup, is:

* no_std <sub>[edit: added; can't believe I missed it, seemed so obvious but thanks Mališa for pointing it out]</sub>
* Provide cryptographic primitives:
  * AEAD encryption/decryption
    * possibly composed of hardware encryption primitives combined with software extension, eg. combining hardware AES with software CCM
    * Ideally streaming AAD
    * Ideally even streaming data (eg. from a scatter-gather buffer), but that might be unrealistic
    * Do we need to split in-place/out-of-place operations? (My focus is on in-place)
  * ECDSA
    * Ideally with streaming AAD
  * ECDH
  * Cryptographic hashes (definitely streaming)
  * HKDF (not sure if extract/expand split is meaningful)
  * Possibly HPKE?
  * Possibly post-quantum mechanisms?
  * Maybe plain symmetric encryption (possibly expressed through COSE's deprecated algorithms mechanism, possibly using an extra mechanism; should have a unified mechanism for group OSCORE)
* Don't impose size limits that the user can not easily tune.
  * Where possible, avoid that the user needs to provide a buffer (through generics or already allocated) for intermediate operations. (👀  ECDSA AAD)
* Query algorithms / curves by COSE number, ideally in a const way.
* Provide associated types that users of the library can use to carry around any intermediate results they need (such that they don't need to know the supported algorithms and thus sizes explicitly).
* Allow backends to only give keys out by handle
* Actually hard decisions:
  * Handle exclusive access? This could be a conflict-heavy point, as users expect that backends to work on a &self (or even a Copy Self, which indeed many backends can provide), whereas some backends need a &mut and can only provide that if they manually and statefully lock themselves, and then enqueue operations (on the stack?) or fail them on reentrancy.
  * Some might want to use eg. signature operations in an async way.
[...]

PR #53, opened 2026-05-27
Add libcrux implementation for AES-GCM (without hardware acceleration)

chrysn:
This does all the busy work without the really juicy bits; based on #50. (Ie. just look at the top-most commit).

Missing are:
- Actual encryption / decryption (because libcrux and embedded-cal disagree on whether in-place or copied is the way to go)
- Tests (boooring ;-) )
- Adding hardware acceleration (won't happen in this PR)

Before we start discussing who can adjust in terms of in-place decryption: Did I maybe just miss how to do it with libcrux? (→ @franziskuskiefer @jschneider-bensch)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,8 +20,14 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
+
+[[package]]
+name = "anyhow"
+version = "1.0.102"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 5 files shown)


jschneider-bensch:
Our APIs currently always need separate plaintext/ciphertext buffers, that's true.

chrysn:
That is rather inconvenient -- of the pilot users (Lakers, libOSCORE and SUIT), the two I know of all operate in-place, and with good reason: An OSCORE implementation's message buffer might easily be the the largest single buffer in the system, not doing AEAD in-place would half the usable message size.

We can hack it in by going back to using `alloc` to make things work initially, but AIU libcrux is just on its way to get rid of that dependency to enable use on constrained systems, and not offering an in-place API would seem to be stopping one exit before being there. Please consider whether an in-place API can be added.

chrysn:
Implemented now using alloc. This has shown a 2nd mismatch: libcrux expects the AAD in linear memory, which is especially not feasible when used for OSCORE. The decision in the design process (https://github.com/lake-rs/embedded-cal/issues/1) was to have streaming AAD here, which is implemented as an impl Iterator.

jschneider-bensch:
We have an offshoot of mainline libcrux which has some experiments (currently just SHA3, ML-KEM, ML-DSA) that are geared towards embedded devices in the [libcrux-iot](https://github.com/cryspen/libcrux-iot) repository.

There I've quickly sketched a version of the mainline `libcrux-aesgcm` crate on branch [`jonas/aes`](https://github.com/cryspen/libcrux-iot/tree/jonas/aes/libcrux-iot/aes) which provides an in-place encryption/decryption API with AAD from an `impl Iterator<Item = &u8>`. The APIs are exposed under `libcrux_iot_aes::aes_gcm_{128, 256}::portable::{encrypt, decrypt}`. Since it's just a rough sketch at the moment, I've taken out the SIMD implementations from mainline libcrux and just done the change for the portable bit-sliced implementation of AES-GCM.

If you could have a look whether those are more in line with the requirements we can think about whether we can implement and maintain these on mainline libcrux. One issue is that they don't fit the current state of the `libcrux_traits` AEAD interfaces that we would like to have as the public API, but we could also think about extending those interfaces.

chrysn:
I've resolved the conflicts by merging back from main. I'll merge this as is; like with the hardware versions, the need to allocate a buffer in some form is tracked in https://github.com/lake-rs/embedded-cal/issues/73, and we'll switch to the branches you mentioned next.

PR #54, opened 2026-05-27
Enhance documentation

chrysn:
Architecture documentation has not been updated since the skeleton has grown some meat around it and needs updating.

<del>(What is currently in the PR is mostly a note not to forget about one aspect; in enhanced docs this would probably be shown visually).</del>

This is now a big "make documentation user-friendly" commit, which (as that is part of how it is presented to the user) entails moving some type around into the `util` module.


chrysn:
I'd like to update a lot of the ALGORITHMS texts to links, and had that briefly in this PR, but as that's super conflict prone, that's best done in a follow-up.

WilliamTakeshi (approved):
LGTM!

chrysn:
Missed following a rename in one place [edit: and clippy caught an erroneous doc-comment that had `///` at module top rather than `//!`]; force-pushed.

Thanks for reviewing, I'll merge as soon as CI passes.

PR #55, opened 2026-05-28
Add HkdfProvider

WilliamTakeshi:
Add HkdfProvider

Obs:
SHA-256 only for now. The blanket impl hardcodes HMAC-SHA-256. But I don't think its hard to make it also works with other hash function.


embedded-cal/src/hkdf.rs
```diff
@@ -0,0 +1,89 @@
+use crate::{HmacAlgorithm, HmacProvider};
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum HkdfError {
+    /// Requested OKM length exceeds 255 × 32 bytes (RFC 5869).
+    OutputTooLong,
+    /// The underlying HMAC produced an output that is not 32 bytes, indicating
+    /// the backend does not correctly implement HMAC-SHA-256 for COSE number 5.
+    InvalidOutputLength,
```
chrysn, line 9: I'm a bit conflicted about whether that should be a user facing error or just an observation of an implementation invariant being violated (which would warrant a panic), but what the hashes do promise is that algorithms produce output of the length they indicate, not that they follow the universally agreed lengths.

Once we have const traits, implementations such as this can put in const assertions that if an algorithm with number N can be constructed, it has that length (and then the runtime error is a violation of the promised API causing a panic, and the disagreement about what the output of any given hash should be is a build time error), but Rust is not there yet.

embedded-cal/src/hkdf.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+pub trait HkdfProvider {
+    /// HKDF-Extract (RFC 5869): returns a 32-byte pseudorandom key.
+    ///
+    /// When `salt` is `None`, 32 zero bytes are used as the HMAC key,
+    /// HashLen = 32 for SHA-256.
+    fn hkdf_extract(&mut self, salt: Option<&[u8]>, ikm: &[u8]) -> Result<[u8; 32], HkdfError>;
```
chrysn, line 17: The extracted material does not need to be visible to the user, so I'd rather hide that in an associated type.

Same for IKM and the output material: Packing it in a type will allow us to create it from DH shared secrets without those secrets ever being user visible.

As for passing in the algorithm, I think we can get away with not having one, but just pass in an HMAC algorithm; HKDF being defined as HMAC based should suffice to justify having those together.

embedded-cal/src/hkdf.rs
```diff
@@ -0,0 +1,89 @@
...
+        ikm: &[u8],
+        info: &[u8],
+        okm: &mut [u8],
+    ) -> Result<(), HkdfError> {
+        let prk = self.hkdf_extract(salt, ikm)?;
+        self.hkdf_expand(&prk, info, okm)
+    }
+}
+
+/// Blanket impl: any [`HmacProvider`] that supports HMAC-SHA-256 (COSE number 5) gets HKDF.
```
chrysn, line 35: A blanket impl rules out that impls do it themselves, and thus also that they keep the secrets out of user memory.

chrysn (approved):
There's a lot of generalization yet to do, but I think that's better handled as incremental improvement -- easier to get things done on the main branch than spread out.

PR #58, opened 2026-06-03
Add hw accelerated for aesccm_16_64_128

WilliamTakeshi:
this PR implements the trait `embedded_cal::AeadProvider` for both `Stm32wba55Cal` and `Nrf54l15Cal`

PR #59, opened 2026-06-03
Add hw accelerated for stm32 aesccm_16_64_256

WilliamTakeshi:
continuation of #58

this PR only implements hardware accelerated for `Stm32wba55Cal`.

Because `CCM` doesn't work for 256 bit keys on Nrf54l15, its not as easy to impl

### DIFF
--- a/embedded-cal-nrf54l15/src/aead.rs
+++ b/embedded-cal-nrf54l15/src/aead.rs
@@ -241,7 +241,12 @@ impl embedded_cal::AeadProvider for super::Nrf54l15Cal {
                 tag.copy_from_slice(&output_buf[message.len()..message.len() + 8]);
                 AeadTag::AesCcm16_64_128(tag)
             }
-            AeadKey::AesCcm16_64_256(_) => todo!(),
+            AeadKey::AesCcm16_64_256(_) => {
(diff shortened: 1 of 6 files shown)


embedded-cal-nrf54l15/src/aead.rs
```diff
@@ -241,7 +241,12 @@ impl embedded_cal::AeadProvider for super::Nrf54l15Cal {
                 tag.copy_from_slice(&output_buf[message.len()..message.len() + 8]);
                 AeadTag::AesCcm16_64_128(tag)
             }
-            AeadKey::AesCcm16_64_256(_) => todo!(),
+            AeadKey::AesCcm16_64_256(_) => {
```
chrysn, line 244: Can at least the AES part be accelerated (albeit then with more software layer on top of it, which we don't have), or not even that?

For if nothing can be done in hardware, then this impl would just not do 16-64-256 and leave it to a software impl, :shrug:.

chrysn: Scratch that, https://github.com/lake-rs/embedded-cal/pull/62 is addressing this.

WilliamTakeshi:
closed because already merged in #62

Issue #57, opened 2026-06-01
Usability pattern: `AeadKey<AeadProvider>`


chrysn:
Around some uses of associated types, I've seen the pattern of providing newtypes or type aliases for associated types that turn an `<_ as AeadProvider>::AeadKey` into `AeadKey<AeadProvider>`.

This wouldn't change anything about our architecture, but I should evaluate whether that can enhance end user usability. (… and then check all the other places where I've used associated types for whether the same applies there).

chrysn:
Exploring the type alias version of this, because it makes extenders much more readable.

chrysn:
The newtype version of this is likely less viable here, because we also need the `&mut Cal`. Those types would probably be done as part of https://github.com/lake-rs/embedded-cal/issues/7.

Issue #51, opened 2026-05-27
Enable RNG trait


chrysn:
Same issue as underlies #49 also exists for the RNG; fixing that once the upcoming PR on empty implementations (useful for testing the libcrux branch) is up.

PR #61, opened 2026-06-03
Add Diffie-Hellman interface

chrysn:
Missing are loading keys from RAM (EC2/OPK and/or non-COSE formats) and everything that comes from making it part of the Cal trait (which is mostly providing all the empty impls); I'll switch this to "ready for review" when those are in as well.

Slightly undecided on whether expressing all the lengths of inputs even makes sense on the algorithm level, as encoded forms might have varying lengths -- and key generation should happen in here anyway (with optional export of keys at generation).


chrysn:
<del>The `SharedSecret` associated type deviates a bit from other output types; I'm PR'ing it in that form for alignment with the `elliptic_curve` crate's trait. Formally, requiring another method call rather than deref'ing to `[u8]` doesn't change *much* for Rust, but passing in the Cal again means that the output type can be a handle that when created doesn't even try to pull the secret out of an enclave, plus it's a good place to document caveats about using shared secrets raw.</del>

That turned out to be a bad idea the way I implemented it originally, as the `&mut` was just to *some* `impl Cal`, not necessarily the own one, so no access to Cal internals through it. Moved back to just opaque type, waiting for #57 to restore its ergonomics.

embedded-cal-rustcrypto/src/dh.rs
```diff
@@ -0,0 +1,209 @@
...
+    type DhAlgorithm = DhAlgorithm;
+    type VisibleSecretKey = VisibleSecretKey;
+    type SecretKey = SecretKey;
+    type PublicKey = PublicKey;
+    type SharedSecret = SharedSecret;
+
+    fn generate_visible(&mut self, alg: Self::DhAlgorithm) -> Option<Self::VisibleSecretKey> {
+        // We're not wrapping anything, so no point in deferring to the self RNG.
+        Some(VisibleSecretKey(match alg {
+            DhAlgorithm::P256 => SecretKey::P256(p256::SecretKey::random(&mut OldRng(self))),
```
WilliamTakeshi, line 15: Do we need the OldRng?

can we do something like:

```rust
DhAlgorithm::X25519 => {
    let mut bytes = zeroize::Zeroizing::new([0u8; 32]);
    self.try_fill_bytes(&mut bytes[..]).ok()?;
    SecretKey::X25519(x25519_dalek::StaticSecret::from(*bytes))
}
```

and use straight from TryRng?

chrysn: There are probably more elegant ways of doing it; I don't know any though (but can look again at whether the `rand_core` crate provides some own compatibility).

Thing is, with X25519 it's easy because AIU, every byte string is a valid key. Generating a P-256 key is more involved, so it makes sense that the random constructor takes an impl Rng – it's just sad that the cryptography traits are lagging so far behind the rest of the ecosystem (in terms of const generic arrays and in terms of the rng trait)..

WilliamTakeshi: I dont know if we can use TryRng directly. Looks like we have to use a loop and keep retrying like:

```rust
DhAlgorithm::P256 => loop {
    let mut bytes = zeroize::Zeroizing::new([0u8; 32]);
    self.try_fill_bytes(&mut bytes[..]).ok()?;
    if let Ok(sk) =
        p256::SecretKey::from_bytes(p256::FieldBytes::from_slice(&bytes[..]))
    {
        break SecretKey::P256(sk);
    }
},
```

Another way is changing `embedded_cal` to have `Rng` instead of `TryRng`

chrysn: The latter is what I'd aim for with https://github.com/lake-rs/embedded-cal/issues/51

WilliamTakeshi (approved):

chrysn:
Merging; notable additions for the commit are:

* RNG through getrandom is added to -rustcrypto. (It stays in this PR because it is not trivial to extract the commit due to entanglement in DH key generation.)

Final rustfmt nits fixed, and rebased using autosquash.

PR #62, opened 2026-06-04
aesccm16_64_256 for nrf54

WilliamTakeshi:
built on top of #59 

nRF54L15 has a CCM00 chip that is dedicated for AES-CCM, but sadly it only works for 128bit key.

For 256bit key, we had to use CRACEN (this PR)

I can see 2 paths from here:

1. continue like this (CCM00 for 128bit key, CRACEN for 256bit keys)
2. unify to use CRACEN for every AES-CCM.

### DIFF
--- a/embedded-cal-nrf54l15/src/aead.rs
+++ b/embedded-cal-nrf54l15/src/aead.rs
@@ -241,7 +241,12 @@ impl embedded_cal::AeadProvider for super::Nrf54l15Cal {
                 tag.copy_from_slice(&output_buf[message.len()..message.len() + 8]);
                 AeadTag::AesCcm16_64_128(tag)
             }
-            AeadKey::AesCcm16_64_256(_) => todo!(),
+            AeadKey::AesCcm16_64_256(_) => {
+                // The CCM00 peripheral is 128-bit only
+                // https://docs.nordicsemi.com/r/bundle/ps_nrf54L15/page/ccm.html-unique_2145434636
+                // https://docs.nordicsemi.com/r/bundle/ps_nrf54L15/page/ccm.html-register.KEY.VALUE
+                todo!()
+            }
         }
     }
 
--- a/embedded-cal-nrf54l15/src/aead.rs
+++ b/embedded-cal-nrf54l15/src/aead.rs
@@ -41,6 +41,53 @@ impl EcbJob {
     }
 }
 
+// ── BA411E AES engine — cryptomaster DMA tags and command words ──────────────────────────────────
+//
+// Sources: sdk-nrf subsys/nrf_security/src/drivers/cracen/sxsymcrypt/src/{cmdma.h,cmaes.h,aead.c}
+//
+// DMATAG layout (input descriptors only; the pusher ignores dmatag on output descriptors):
+//   bits [3:0]  — engine selector  (DMATAG_BA411 = 1)
+//   bit    [4]  — config-register write flag  (DMATAG_CONFIG sets this)
+//   bit    [6]  — header/AAD data type flag   (DMATAG_DATATYPE_HEADER)
+//   bits [12:8] — trailing padding bytes to ignore  (DMATAG_IGN, see descriptor::dmatag_ign)
+//   bits [15:8] — config register offset when bit 4 is set
+
+const DMATAG_BA411: u32 = 1;
+
+// Config-register descriptors (bit 4 set; register offset in bits [15:8]):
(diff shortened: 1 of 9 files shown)


malishav:
Any idea on the performance difference between CCM00 and CRACEN?

WilliamTakeshi:
> Any idea on the performance difference between CCM00 and CRACEN?

Did a fast check  (on https://github.com/WilliamTakeshi/embedded-cal/tree/aesccm_16_64_256_nrf_benchmark). looks like CCM00 is 27–30% better.

Using CCM00:
AesCcm16_64_128 encrypt 64B: 25572 cycles/op
AesCcm16_64_128 decrypt 64B: 29415 cycles/op


Using CRACEN:
AesCcm16_64_128 encrypt 64B: 36792 cycles/op
AesCcm16_64_128 decrypt 64B: 40294 cycles/op

chrysn:
I'm not the one who has to present numbers on being the fastest, but I suggest going with CRACEN at least if 3 or 4 holds:
1. general simplicity of maintenance
2. there'll be more hardware that has CRACEN coming up (nRF92 was announced with it, IIRC)
3. it looks like CRACEN is a bit more flexible when it comes to scatter-gather data (DescriptorChain has an N we could thread through further to effectively make the caller say how many parts there need to be)
4. CRACEN's buffer look more flexible when it comes to using the same memory as input and output buffer

(and maybe we can also realize code size benefits by reusing the same code for both lengths)

For the extra 30% there can still be a fast path for cases when, say, the AAD is small and/or unscattered.

embedded-cal-nrf54l15/src/aead.rs
```diff
@@ -1,35 +1,312 @@
-// FIXME: Something is probably there, just not implemented yet.
+// ── BA411E AES engine — cryptomaster DMA tags and command words ──────────────────────────────────
```
chrysn: minor:

```suggestion
// # BA411E AES engine — cryptomaster DMA tags and command words
```

… and as that comment appears to apply to all in here, it might make sense to pack them into an inline module and make this comment apply to the module.

embedded-cal-nrf54l15/src/aead.rs
```diff
@@ -1,35 +1,312 @@
...
+    fn as_ref(&self) -> &[u8] {
+        match self {
+            AeadTag::AesCcm16_64_128(r) => &r[..],
+            AeadTag::AesCcm16_64_256(r) => &r[..],
+        }
+    }
+}
+
+impl super::Nrf54l15Cal {
+    fn ccm_encrypt<const KEY_LEN: usize>(
```
chrysn, line 94: We should check but can do this in cleanup:

Does monomorphization cause this function to be duplicated here? If so, could split into an inlineable outer part that just does the const assert, and an inner part that works on a key slice.

(Same for decrypt; actually, there might be so much overlap to even consider running both through the same workhorse function.)

embedded-cal-nrf54l15/tests/integration.rs
```diff
@@ -20,13 +21,14 @@ mod tests {
 
     #[init]
     fn init() -> super::TestState {
+        let raw = embedded_cal_nrf54l15::Nrf54l15Cal::new(nrf_pac::CRACEN_S, nrf_pac::CRACENCORE_S);
```
chrysn, line 24: Why two, and what's the difference between raw and base?

embedded-cal/src/aead.rs
```diff
@@ -125,6 +125,22 @@ impl AadGenerator for &[&[u8]] {
     }
 }
 
+/// Build the CCM B0 block as defined in RFC 3610.
+///
+/// B0 = flags | nonce | Q, where Q is the message length encoded in L bytes.
+pub fn build_b0(nonce: &[u8], msg_len: usize, a_len: usize, tag_len: usize) -> [u8; 16] {
```
chrysn, line 131: I think we should (but not necessarily right now) move this somewhere off the most public path. We don't have finished AES plumbing yet, but this could already go into an AES plumbing module. (It's not plumbing precisely as we have it for SHA, because here it's a tool for an impl to use rather than an interface that gets driven, but still, that's the closest we have).

chrysn (approved):
I've reviewed this as a whole, thus including #58 and #59.

Looks sensible enough for me to merge; please either merge as is (then the comments go into cleanup tracking) or address any you'd prefer to do now and merge then.

malishav:
@WilliamTakeshi could you do a quick test on larger payloads? e.g. 128 and 256 bytes. I am curious whether there is a one-off CRACEN access price to pay that would become less important with larger payloads.

PR #69, opened 2026-06-09
Hkdf enhancement

WilliamTakeshi:
Some points raised in the #55

It changes the HKDF to not only work with HMAC-SHA256, but with any HMAC.
Improve the comments a bit

### DIFF
--- a/embedded-cal/src/hkdf.rs
+++ b/embedded-cal/src/hkdf.rs
@@ -2,71 +2,105 @@ use crate::{HmacAlgorithm, HmacProvider};
 
 #[derive(Debug, PartialEq, Eq)]
 pub enum HkdfError {
-    /// Requested OKM length exceeds 255 × 32 bytes (RFC 5869).
+    /// Requested OKM length exceeds 255 × HashLen bytes (RFC 5869).
     OutputTooLong,
-    /// The underlying HMAC produced an output that is not 32 bytes, indicating
-    /// the backend does not correctly implement HMAC-SHA-256 for COSE number 5.
+    /// The underlying HMAC produced an unexpected output length.
     InvalidOutputLength,
 }
 
 pub trait HkdfProvider {
-    /// HKDF-Extract (RFC 5869): returns a 32-byte pseudorandom key.
+    type Algorithm: HmacAlgorithm;
+    /// Opaque input keying material. May be constructed from a DH shared secret
+    /// without the raw bytes ever being user-visible.
+    type Ikm: ?Sized;
+    /// Opaque pseudorandom key — output of Extract, input to Expand.
+    type Prk;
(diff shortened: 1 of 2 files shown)


embedded-cal/src/hkdf.rs
```diff
@@ -2,85 +2,121 @@ use crate::{HmacAlgorithm, HmacProvider};
...
+
+    fn hkdf_extract(
+        &mut self,
+        alg: Self::Algorithm,
+        salt: Option<&[u8]>,
+        ikm: &[u8],
+    ) -> Result<Self::Prk, HkdfError> {
+        // When salt is absent, RFC 5869 uses HashLen zero bytes as the HMAC key.
+        // Buffer covers standard algorithms up to SHA-512 (64 bytes).
+        // Ideally this would be H::Algorithm::MAX_OUTPUT_LEN once const_trait_impl stabilises.
```
chrysn, line 70: We don't need const_trait_impl for that if it is an associated const -- we can just introduce that associated constant.

Then InvalidOutputLength is back to returning inconsistent lengths, and as we start just using the underlying type as the only truth (no longer adding own knowledge), I don't think there are any cases left for it.

embedded-cal/src/hkdf.rs
```diff
@@ -2,85 +2,121 @@ use crate::{HmacAlgorithm, HmacProvider};
...
-    /// The underlying HMAC produced an output that is not 32 bytes, indicating
-    /// the backend does not correctly implement HMAC-SHA-256 for COSE number 5.
+    /// The algorithm's output length exceeds the supported maximum (64 bytes), or the underlying
+    /// HMAC returned a length inconsistent with the algorithm's own reported length.
     InvalidOutputLength,
 }
 
 pub trait HkdfProvider {
-    /// HKDF-Extract (RFC 5869): returns a 32-byte pseudorandom key.
+    type Algorithm: HmacAlgorithm;
```
chrysn, line 13: Can't we just say that `trait HkdfProvider: HmacProvider`, and then not use an own Algorithm at all but just everywhere use `<Self as HmacProvider>::Algorithm`?

embedded-cal/src/hkdf.rs
```diff
@@ -2,85 +2,121 @@ use crate::{HmacAlgorithm, HmacProvider};
...
+    /// Opaque input keying material. May be constructed from a DH shared secret
+    /// without the raw bytes ever being user-visible.
+    type Ikm: ?Sized;
```
chrysn, line 16: Not having any `From<[u8]>` for Ikm and `AsRef<[u8]>` for Okm means that users depend on knowing the concrete type -- which they can because of the blanket implementation, but then, we could just also not have types here until we start differentiating. (And that really doesn't need to be now.)

Issue #46, opened 2026-05-25
Document or revisit fallibility


chrysn:
The current APIs are designed to only be fallible where failure is expected and recoverable at runtime -- e.g. when querying whether an algorithm is supported, or a key is loaded that fails some checks.

There has [been some discussion](https://matrix.to/#/!HKnrmXbIiYsaCwImIV:matrix.org/$lfQrhwCqirH33Xotet4KvXqSrc1Olvij--LCJQ-cDs4?via=matrix.org&via=mozilla.org) on whether that is the right choice for all situations; in particular, a hardware accelerator connected via some bus might simply report a bus error.

The current approach is to treat those like irrecoverable hardware errors (just like a memory access fault due to a bad RAM cell would result in a crash) after which no sensible device operation is possible, and turn such failures into panics, rather than lugging around an error state that most implementations do not even inhabit. (Granted, those most implementations could use the `Infallible` type there).

I still think that that is the right path, but it should be documented -- and if there's any reason to change something about it, this issue is a good place to make any point about it.

chrysn:
One option is to have an associated RuntimeError type, which implementers are encouraged to use Infallible for. Then some operations (including Rng?) would have `Result<T, Self::RuntimeError>` which users might unwrap (or even match on to make the infallibility clear if they require that RuntimeError is infallible) or pass on.

Issue #40, opened 2026-05-25
Move traits from dependencies into methods


chrysn:
We're racking up a large number of dependent traits in Cal -- that is expected and part of the design, but impractical to implement (roughly adding 20 lines per impl per trait if that trait is unused) and not user friendly (the traits might easily have overlapping methods, and then all of a sudden you find yourself writing `<MyCal as HmacProvider>::update(mycal, …)` and that's ugly).

An possible alternative is for the base Cal to just have associated types for each of the traits:

```rust
struct Cal {
    type HashProvider: HashProvider;
    type HmacProvider: HmacProvider;
    …

    fn as_hmac_provider(&mut self) -> &mut Self::HmacProvider;
}
```

In many cases, the HmacProvider will need to reference the HashProvider internally, so for anything but pure software implemntations, the sensible impl is this:

```rust
struct Cal {
    type HashProvider = Self;
    type HmacProvider = Self;
    …

    fn as_hmac_provider(&mut self) -> &mut Self {
        self
    }
}
```

but this cuts down the boilerplate to 4 lines per trait, and the `NoHmacAlgorithm` and similar uninhabited types can just grow a companion ZST for `UselessHmacProvider` (TBD better name), which does the 20line empty impl once. Alternatively, all those could be fallible (`-> Option<Self::HashProvider>` allowing having an uninhabited type there), but as all our types already have some form of fallibility internally (through the algorithm selection where no algorithm can be supported, or through the RNG being TryRng), that'd just create two error cases for the user to handle, for little benefit.

## When

Not now; we have many branches in the air, and this is best done as a flag day operation, possibly with some renaming (as name collisions then hit less hard).

chrysn:
Working on this now that no other PRs are pending.

One benefit to best reap immediately after the heavy changes is renaming the associated Algorithm types (because what they're an algorithm for is now in the path).

Some of #57 may make sense now too, as `<C::DhProvider as DhProvider>::DhAlgorithm` (later s/DhAl/Al/, but still) is becoming a mouthful

chrysn:
One funny thing about the PR is that I found that some libcrux impls were empty rather than forwarding. This is now much more visible, and being fixed right in the PR.

PR #71, opened 2026-06-10
RNG: Expect (but do not require) infallible RNG

chrysn:
Threading the option of hardware failure all the way through would create a lot of application code that doesn't know how to deal with it -- and why should it, "hardware isn't doing what it should" is irrecoverable and panic-worthy unless we decide differently in https://github.com/lake-rs/embedded-cal/issues/46.

This turns all the cases where we touch RNGs into infallible RNGs. (Note that the impl is still of `TryRng`, as `Rng` has a blanket impl on `Twhere T: TryRng<Error = Infallible>` and thus can and should not be implemented).

It does not yet act on #51 (make Rng a dependency of Cal), because so far things work fine that way (and #40 would make short work of any changes around there anyway).

### DIFF
--- a/embedded-cal-nrf54l15/src/try_rng.rs
+++ b/embedded-cal-nrf54l15/src/try_rng.rs
@@ -2,29 +2,9 @@ use crate::Nrf54l15Cal;
 use nrf_pac::cracencore::vals::{ControlSoftrst, State};
 const MAX_TRNG_RESTARTS: u32 = 3;
 
-/// Error returned by [`RngProvider::try_fill_bytes`].
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum RngError {
-    /// The hardware noise source failed its internal health test too many times.
-    ///
-    /// This indicates the entropy source may be untrustworthy (degraded oscillator,
-    /// power instability, or a silicon fault).
-    HardwareFailure,
-}
-
-impl core::fmt::Display for RngError {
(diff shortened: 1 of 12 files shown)


embedded-cal-nrf54l15/src/try_rng.rs
WilliamTakeshi: update comment

embedded-cal-stm32wba55/src/try_rng.rs
```diff
@@ -57,10 +37,10 @@ impl rand_core::TryRng for Stm32wba55Cal {
                 // spinning forever or handing out potentially non-random bytes.
```
WilliamTakeshi: update comment

embedded-cal/src/empty.rs
```diff
@@ -156,24 +153,6 @@ impl<const PLUMBING: bool> DhProvider for EmptyCal<PLUMBING> {
     }
 }
 
-impl<const PLUMBING: bool> rand_core::TryCryptoRng for EmptyCal<PLUMBING> {}
-
-impl<const PLUMBING: bool> rand_core::TryRng for EmptyCal<PLUMBING> {
-    type Error = NoRng;
```
WilliamTakeshi, line 162: since we are not using the error anymore, we can delete `pub struct NoRng;` and related code

WilliamTakeshi (approved):
Everything looks good, just nitpicked some comments to be updated and a dead code

PR #74, opened 2026-06-11
CI: Disable F* attempts, rename main task

chrysn:
Work in hax is moving from F* to Lean, so we will likely not re-enable (or pass) the current hax CI task precisely as is.


chrysn:
Updated branch rules so the new name is recognized. Remaining PRs that get done before trait refactoring flag day (today!) will need this merged in, but the one where it applies (https://github.com/lake-rs/embedded-cal/pull/53) needs that anyway.

PR #76, opened 2026-06-11
Clean up rustcrypto and turn it into an extender

chrysn:
Cleaning up after https://github.com/lake-rs/embedded-cal/pull/75 I found things off around the -rustcrypto trait; in particular, it didn't do HMAC, and worse, it did it manually rather than using #75.

… when really, a) it should provide HMAC based on its hash algorithms, and b) it doesn't need to be limited to implemented Rustcrypto hashes here because especially the HMAC could compose onto other components.

This PR's scope is "whatever I can do quickly to make this better using state in hot mental cache before I should go back to signatures".


chrysn:
One thing I found out during this is that for VisibleSecretKey->SecretKey, I picked the wrong one of From/Into, hampering building rather generic wrappers. But that can be rectified, and thus this exercise is useful in honing the ECDH traits before doing something similar with ECDSA.

chrysn:
This started out as fixing HMAC/HKDF on rustcrypto -- but there's a bit more to do there, so I'll leave it for review as is.

(HMAC/HKDF will need more to accommodate the EDHOC idiom of prescribing that only HKDF-expand is used, and that KMAC is used in other cases, all depending on the hash algorithm which is really the input here.)

embedded-cal-rustcrypto/src/dh.rs
```diff
@@ -130,58 +149,82 @@ impl embedded_cal::DhProvider for RustcryptoCal {
...
 }
 
-const MAX_SHARED_SECRET_LENGTH: usize = 32;
-
-#[derive(Zeroize, ZeroizeOnDrop)]
-pub struct SharedSecret(heapless::Vec<u8, MAX_SHARED_SECRET_LENGTH>);
+pub enum SharedSecret<BSS> {
+    Length32([u8; 32]),
+    Direct(BSS),
+}
```
WilliamTakeshi, line 230: is there a reason to remove zeroize?

WilliamTakeshi (approved):

embedded-cal-rustcrypto/src/dh.rs
chrysn: Deriving Zeroize would require having Zeroize as a requirement on the underlying BSS, and so far we don't provide that.

PR #78, opened 2026-06-12
fix compilation error

WilliamTakeshi:
#75 The tests for both boards no longer compile. This PR fixes the issue.


chrysn:
Thanks. Could you take a moment to check whether we can make CI catch those next time?

PR #84, opened 2026-06-15
Implement DhProvider for STM32 and nRF54

WilliamTakeshi:
This PR implements
-  P256 for STM32
-  P256 + X25519 + X448  for nRF54
- Adds Zeroize and try to clear memory places where its needed 

Some `fn` are duplicated in both files, (ge, sub256, add_mod, mul_mod, reduce_p256, pow_mod, bytes_to_words, words_to_bytes).

On these cases should we create a new crate to import from both sides?


chrysn:
I think that placing the functions in a dh plumbing module (similar to build_b0) would make sense for the moment. The introduction of (or EC) plumbing should remove the need for them, but until then, this should do.

WilliamTakeshi:
Deduplicated functions on  [`87b9798`](https://github.com/lake-rs/embedded-cal/pull/84/commits/87b9798999c71f70499a4070b7974aa4b0b38078)

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+    CRACEN_PKE_RAM_BASE + slot * SLOT_SIZE
+}
+
+fn pke_ram_word(addr: u32) -> Reg<u32, RW> {
+    // Safety: addr is always a valid CRACEN PKE RAM address derived from CRACEN_PKE_RAM_BASE.
+    unsafe { Reg::from_ptr(addr as *mut u32) }
+}
+
+fn write_pke_le(addr: u32, data: &[u8]) {
+    for (i, chunk) in data.chunks_exact(4).enumerate() {
```
chrysn:
```suggestion
fn write_pke_le(addr: u32, data: &[u8]) {
    debug_assert!(data.len() % 4 == 0, "function expects word sized data");
    for (i, chunk) in data.chunks_exact(4).enumerate() {
```

(in a sense that's just as much for documentation than to catch bugs)

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+// Little-endian slot address (Montgomery curves): data at start of slot.
+#[inline(always)]
+fn montgomery_slot_addr(slot: u32) -> u32 {
+    CRACEN_PKE_RAM_BASE + slot * SLOT_SIZE
+}
+
+fn pke_ram_word(addr: u32) -> Reg<u32, RW> {
+    // Safety: addr is always a valid CRACEN PKE RAM address derived from CRACEN_PKE_RAM_BASE.
```
chrysn: If that function is to be safe, it needs to check those constraints.

Otherwise, seeing that it is always called with arguments of the shape `k + 4 * d`, maybe separate those args and assert on k and d separately?

If the asserts are an issue for performance, either this function would need to be unsafe, or the k and d could have newtypes that wrap the promises.

(I guess that for an MVP, an assert on the addr being in the right range would be most practical.)

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
+use embedded_cal::DhAlgorithm as _;
+use embedded_cal::p256::{P256_ORDER, bytes_to_words, ge, p256_recover_y};
+use nrf_pac::common::{RW, Reg};
+use nrf_pac::cracencore::vals::{Selcurve, Swapbytes};
+use rand_core::Rng as _;
+use zeroize::{Zeroize, ZeroizeOnDrop};
+
+const CRACEN_PKE_RAM_BASE: u32 = 0x5180_8000;
```
chrysn, line 14: Please give some source for numbers such as this.

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+}
+
+fn read_pke_le(addr: u32, out: &mut [u8]) {
+    for (i, chunk) in out.chunks_exact_mut(4).enumerate() {
+        chunk.copy_from_slice(&pke_ram_word(addr + i as u32 * 4).read().to_le_bytes());
+    }
+}
+
+// P-256 generator point (big-endian bytes, for use with cracen_ecc_mult).
+const P256_GX: [u8; 32] = [
```
chrysn: If we could have this constant in the P256 plumbing easily, that'd be helpful for later refactoring towards full plumbing. (But if that's not easy, eg. because it'd need complex const conversion, fine for an MVP).

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+    alg: DhAlgorithm,
+    // First `alg.output_length()` bytes are valid.
+    x: [u8; 56],
+    // Used only for P-256 (recovered on import).
+    y: [u8; 32],
+}
+
+#[derive(Zeroize, ZeroizeOnDrop)]
+pub struct SharedSecret {
+    bytes: [u8; 56],
```
chrysn, line 180: At some point in time we'll probably want to come up with a style guide that helps implementers decide whether the preferred way to do this is to have an enum with algorithms, this, or a heapless Vec -- and which of those are preferred for which type and why.

For the time being, we probably goes with what is easiest at that moment, and collect experience.

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+
+        let status = self.cracen_core.pk().status().read();
+        debug_assert!(
+            status.errorflags() == 0 && status.failptr() == 0,
+            "CRACEN PKE scalar multiplication failed (errorflags={:#x}, failptr={:#x})",
+            status.errorflags(),
+            status.failptr(),
+        );
+
+        let mut rx = [0u8; 32];
```
chrysn, line 238: Just a question for the next step of writing plumbing: Ideally we'd not do this and just leave the data in the slots during larger operations. Does the choice of slot (here: of SLOT_RESULT) matter, or could the Cracen Cal plumbing describe that it has N slots (a bit like general-purpose registers) that can be used before they need to be spilled on the stack?

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+    len: usize,
+}
+
+impl super::Nrf54l15Cal {
+    fn wait_pk_ready(&mut self) {
+        while self.cracen_core.pk().status().read().pkbusy() {}
+        while self.cracen_core.ikg().status().read().ctrdrbgbusy() {}
+    }
+
+    pub(super) fn cracen_ecc_mult(
```
chrysn: Please document for easier comparison with `cracen_montgomery_mult`.

embedded-cal-nrf54l15/src/dh.rs
```diff
@@ -0,0 +1,466 @@
...
+        &mut self,
+        alg: Self::Algorithm,
+        secret: &[u8],
+    ) -> Result<Self::VisibleSecretKey, embedded_cal::ImportError> {
+        let expected = alg.output_length();
+        if secret.len() != expected {
+            return Err(embedded_cal::ImportError);
+        }
+        let mut scalar = [0u8; 56];
+        scalar[..expected].copy_from_slice(secret);
```
chrysn, line 384: If an attacker can feed us a private key, in a sense we're already lost -- but still, I'd think that validation is prudent here, even if only to guard against accidental misuse.

embedded-cal-nrf54l15/src/microcode.rs
```diff
@@ -0,0 +1,169 @@
+// CRACEN PKE microcode (4892 bytes = 1223 u32 words).
+// Source: sdk-nrf subsys/nrf_security/src/drivers/cracen/cracenpsa/src/microcode_binary.h
```
chrysn: This is not licensed compatibly to this crate.

(Insert sad pikachu face here).

I think the best fix is to
- add a note in SPDX format in here
- Change the license of the embedded-cal-nrf54l15 crate to include that license

If you're not familiar with those, I can run that.

chrysn (commented):
First pass through the nRF54L15.

embedded-cal-nrf54l15/src/dh.rs
chrysn: Tracked in https://github.com/lake-rs/embedded-cal/issues/85

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: On nrf54 the SLOT doesn't matter. We are the ones choosing at `cracen_core.pk().pointers()`.

The only limitation seems like we have a limited choice of SLOTS (we can choose any between `6..=13`)

They usually are divided in pairs, and called `PTR_A/B/C/AA`. So yeah, I think we can have a more generic `fn` that doesn't clear the slots, just keep the swapping the input and output.

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: updated to cracen_p256_mult and added some comments on [`598ec12`](https://github.com/lake-rs/embedded-cal/pull/84/commits/598ec1297e845e8dfee281f4c822eb89ba8e32da)

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: added comments here: https://github.com/lake-rs/embedded-cal/pull/84/commits/eeedb849c00e316ccf3ee492471e46e9f9f6d837

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: added 2 checks, one for if the pke_ram_word is inside the pke_data 
and another for alignment

[`eeedb84`](https://github.com/lake-rs/embedded-cal/pull/84/commits/eeedb849c00e316ccf3ee492471e46e9f9f6d837)

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: needed to update a bit (use while instead of forloops) but seems like its nicer.

[`1f74aea`](https://github.com/lake-rs/embedded-cal/pull/84/commits/1f74aea3c99b064e4b6d2133727c2843a4e4e547)

embedded-cal-nrf54l15/src/microcode.rs
WilliamTakeshi: we need our license before that? I am not familiar with licenses, so I don't know how to do it

WilliamTakeshi:
I think its ready for review again!

embedded-cal-nrf54l15/src/microcode.rs
chrysn: I've talked with NLnet on what we can do there; once https://github.com/lake-rs/embedded-cal/pull/86 is in, we can rebase this onto that, and

* This file (which is ideally stripped to really the bare data, not the loading tool) gets a diverging note replicating the SPDX license and copyright of the original file <https://github.com/nrfconnect/sdk-nrf/blob/500e9fa603935c12c852df090a5c1354fb8dd904/subsys/nrf_security/src/drivers/cracen/common/src/cracen/hardware/microcode_binary.h#L4>.
* The license gets copied into the LICENSES folder (the reuse.software check complains otherwise anyway)
* The crate gets a "(MIT OR Apache-2.0) AND LicenseRef-Nordic-5-Clause" license
* README gets a note that some hardware backends depend on microcode and thus have a more restrictive license (but all the software written in this project is MIT/Apache).

embedded-cal-nrf54l15/src/dh.rs
chrysn: Thanks -- but as it upholds a soundness constraint, it either needs to be a full assert or the function stays unsafe. (In the latter case, docs can say on the function that "Address needs to be in the right range; all safety properties are checked in debug mode" and the users can say that "Addresses are in ranged, as checked in debug mode").

embedded-cal/src/p256.rs
```diff
@@ -0,0 +1,232 @@
+// P-256 field arithmetic shared by hardware DH back-ends.
+//
+// All [u32; 8] arrays use little-endian word order: index 0 is the least-significant word.
```
chrysn, line 6: I think there's a lot of improvement that *could* be made around using types and stuff, but let's leave it like it is, because a next step is pulling this into a plumbing layer anyway, and things would be torn apart then again anyway.

(Let's leave this as unresolved so I can pick up on it, but no need for change before merging.)

embedded-cal-nrf54l15/src/dh.rs
chrysn: Thanks. (Leaving this open for visibility so I can pick it up later, but it doesn't block merging).

chrysn (commented):
I've read through all of it: Thanks, it's looking good, see the open comments (which realistically mean rebase onto #86) and then we're good to go.

embedded-cal-nrf54l15/src/microcode.rs
chrysn: If you prefer I can also do that -- in that case, let me know when you're done with the rest, and I'll do the relevant merges and added annotations.

embedded-cal-nrf54l15/src/dh.rs
WilliamTakeshi: Yes, you are right!

Updated on [`1966a19` (this PR)](https://github.com/lake-rs/embedded-cal/pull/84/commits/1966a19d6afdad379341a1c756c366963175ccb9) where we propagate the unsafe to the write/read functions, and write safety messages on each usage

WilliamTakeshi:
@chrysn rebased and addressed the unsafe thing. Also added SPDX on all new files except for microcode. Can you do the SPDX of the microcode?

PR #86, opened 2026-06-15
Update license to MIT/Apache-2.0, and set SPDX headers

### LINKED ISSUE #63: Add license
We are missing a license in the repository. We agreed to license the code under a license recognized by OSI and FSF. We always used 3-clause BSD in our projects but I don't have much experience with the nuances of different options. Any objections on adopting 3-clause BSD?

@chrysn @franziskuskiefer

chrysn:
Closes: https://github.com/lake-rs/embedded-cal/issues/63

Current status is incomplete; copyright holders still need to be set, and CI needs to be set up.


chrysn:
I did start initially to look into who authored what, but that'd have been complicated to take Co-Authored-By into account, so I just put everyone everywhere. I've taken the credits from the README; let me know if you prefer individual authors' copyright notes (but to my knowledge, all who contributed are hired for this, which makes the copyright note usually a company one; it's not precise authorship as captured in git).

Speaking of git, I've added a .mailmap to make git authorship consistent (and having that is good and correct even if we don't use it to generate copyright headers).

WilliamTakeshi (approved):

PR #87, opened 2026-06-24
Plumbing for ECC

chrysn:


chrysn:
First stage is approaching review readiness; still open:

1. Plumbing API for accelerated scalar operations (that's why the plumbing doesn't have a SUPPORTED but HAS_MULTIPLY_SCALAR_POINT; those will be per-function, effectively) – not needed for MVP though.
2. <del>Loading and storing from bytes representations is missing (that's why the nRF is not a software-only high-level shim yet)</del> – basics are there, but cutting that cord best works after the refactoring in 3.
3. Some refactoring is better done later as it is about moving between files (and that's hard to review unless it's in stand-alone commits or even PRs).

chrysn:
I think this is now ready for a first round of review; @WilliamTakeshi, I'd appreciate a test on your hardware as a first step.

@franziskuskiefer / @jschneider-bensch, is that a minimal workable interface for you?

WilliamTakeshi:
Checked against hardware on `9259a875d`, everything pass. I will take a deeper look later

```
(1/9) running `test_hash_algorithm_sha256`...
(2/9) running `test_hmac_sha256`...
(3/9) running `test_hkdf_sha256`...
(4/9) running `test_tryrng`...
(5/9) running `test_aead_aesccm_16_64_128`...
(6/9) running `test_aead_aesccm_16_64_256`...
(7/9) running `test_dh_ecdh_p256`...
(8/9) running `test_dh_x25519`...
(9/9) running `test_dh_x448`...
all tests passed!
Firmware exited successfully
```

embedded-cal/src/error.rs
```diff
@@ -0,0 +1,18 @@
...
+#[derive(Debug)]
+pub struct ImportError;
+
+impl core::fmt::Display for ImportError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str("data not valid for algorithm")
+    }
+}
+
+impl core::error::Error for ImportError {}
```
jschneider-bensch, line 18: Do we have any requirements regarding the minimum supported Rust version? E.g. I think `core::error::Error` is stable since 1.81.

(I don't know what the current MSRV is of embedded-cal, or of its intended primary users, and I don't know if we have to think too much about it. Just wondering if there are some thoughts around it, when I saw the `impl` here again.)

embedded-cal/src/plumbing/ec.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// We can't pull in an associated constant to make it take an array (and depending on the
+    /// invariants needed for the curves beyond clamping, it would then stillb e a faillible
+    /// operation).
+    ///
+    /// One option for getting earlier errors would be to take a reference to a length-generic
+    /// array and const-assert on the lengths (causing not a type but at least a conditional
+    /// compilation error), but that will need some testing w/rt ergonomics.
+    fn import_scalar_bytes(&mut self, scalar: &[u8]) -> Result<Self::Scalar, crate::ImportError>;
+
+    fn point(&mut self, x: Self::Scalar, y: Self::Scalar) -> Self::Point;
```
jschneider-bensch, line 80: As I understand the signature, this assembles a `Point` from an affine representation as two scalars. Should it do any kind of validation of this affine representation, i.e. be fallible?

jschneider-bensch (commented):
I think we can try to implement an abstraction layer based on this in `libcrux-p256`. We'd need to implement the high-level operations in `libcrux-p256` generically over an implementor of `ec::plumbing::EcPrimitives<P256>` and also implement `ec::plumbing::EcPrimitives<P256>` using the existing libcrux implementation of the functionality in the trait, right?

I'm not sure I understand at the moment how we'd extend `embedded-cal-libcrux` based on that, though.

embedded-cal/src/error.rs
chrysn: See https://github.com/lake-rs/embedded-cal/pull/98

embedded-cal/src/plumbing/ec.rs
chrysn: The way it is used here so far is with the caller's assertion that this is a point on the curve – as the input comes from point reconstruction.

So at this stage I'll fix this in the docs to say that the point is required to be on the curve (and it's the caller's responsibility to check that – after all, this is the plumbing and it's there to just give access to the accelerator).

(Also, now I remember why that function is not documented – because that question was not answered.)

chrysn:
Had to rebase because changes to the libcrux AEAD parts conflicted with a change (and it'd have taken a rebase anyway due to the fixups).

WilliamTakeshi (approved):
LGTM

PR #89, opened 2026-06-25
sha3 for nRF54L15

WilliamTakeshi:
Looking deeper on the CRACEN documentation, we have access to SHA3 (just on another module `BA418`)

So this PR implements directly the SHA3 (224, 256, 384, 512) for nRF54 (there is no SHA3 in the STM32WBA55 :/)

That means that things like padding is directly baked on the `finalize`, etc. I didn't create an plumbing for it, since I don't know what to make generic for.

### DIFF
--- a/embedded-cal-nrf54l15/src/sha3.rs
+++ b/embedded-cal-nrf54l15/src/sha3.rs
@@ -1,0 +1,225 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-FileCopyrightText: Inria-AIO, Cryspen, and Christian Amsüss
+
+use crate::descriptor::{DescriptorChain, Input, Output};
+
+// BA418 (SHA-3) engine mode codes.
+#[repr(u8)]
+#[derive(PartialEq, Eq, Debug, Clone, Copy)]
+enum Variant {
+    Sha3_224 = 0x06,
+    Sha3_256 = 0x07,
+    Sha3_384 = 0x0B,
+    Sha3_512 = 0x0F,
+}
+
+impl Variant {
+    const fn output_len(self) -> usize {
+        match self {
+            Variant::Sha3_224 => 28,
+            Variant::Sha3_256 => 32,
+            Variant::Sha3_384 => 48,
+            Variant::Sha3_512 => 64,
+        }
+    }
+
+    // Keccak absorption rate in bytes: (1600 - 2 * digest_bits) / 8.
+    const fn rate(self) -> usize {
+        match self {
+            Variant::Sha3_224 => 144,
+            Variant::Sha3_256 => 136,
+            Variant::Sha3_384 => 104,
+            Variant::Sha3_512 => 72,
+        }
+    }
+}
+
+// Largest Keccak rate across all SHA-3 variants (SHA3-224 = 144 bytes).
+const MAX_RATE: usize = 144;
+// Full Keccak-f[1600] permutation state (1600 bits).
+const KECCAK_STATE_SIZE: usize = 200;
+
+// BA418 DMA tags.
+// DMATAG_BA418 = 5, DMATAG_CONFIG = 16.
+// DMATAG_LAST (bit 5 = 32) is ORed in automatically by DescriptorChain on the
+// final descriptor of each chain, so data descriptors start at 5.
+const DMATAG_BA418_CONFIG: u32 = 5 | 16; // = 21
+const DMATAG_BA418_DATA: u32 = 5;
+// DMATAG_BA418 | DMATAG_DATATYPE(1) | DMATAG_LAST: re-injects a saved Keccak sponge
+// state. DMATAG_LAST here is a BA418 semantic flag (not the chain terminator): it
+// marks this descriptor as a state-load operation.
+const DMATAG_BA418_INITIAL_STATE: u32 = 5 | (1 << 6) | (1 << 5); // = 101
+
+// SHA3_SAVE_CONTEXT (bit 6 of the BA418 config word): when set, the hardware outputs
+// the 200-byte intermediate Keccak sponge state instead of the final digest.
+const SHA3_SAVE_CONTEXT: u8 = 1 << 6;
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum HashAlgorithm {
+    Sha3_224,
+    Sha3_256,
+    Sha3_384,
+    Sha3_512,
+}
+
+impl HashAlgorithm {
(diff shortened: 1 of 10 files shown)


embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+            Variant::Sha3_224 => 144,
+            Variant::Sha3_256 => 136,
+            Variant::Sha3_384 => 104,
+            Variant::Sha3_512 => 72,
+        }
+    }
+}
+
+// Largest Keccak rate across all SHA-3 variants (SHA3-224 = 144 bytes).
+const MAX_RATE: usize = 144;
```
chrysn, line 38:
```suggestion
// Largest Keccak rate across all SHA-3 variants
const MAX_RATE: usize = Variant::Sha3_512::rate();
```

(Why comment when code can tell, and fewer magic numbers is good.)

embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+        match self {
+            Variant::Sha3_224 => 28,
+            Variant::Sha3_256 => 32,
+            Variant::Sha3_384 => 48,
+            Variant::Sha3_512 => 64,
+        }
+    }
+
+    // Keccak absorption rate in bytes: (1600 - 2 * digest_bits) / 8.
+    const fn rate(self) -> usize {
```
chrysn, line 27: I've played around a bit with putting that formula into a debug_assert rather than a comment, but that always made the code less readable, and it's not like those numbers could ever be so different.

embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+// SHA3_SAVE_CONTEXT (bit 6 of the BA418 config word): when set, the hardware outputs
+// the 200-byte intermediate Keccak sponge state instead of the final digest.
+const SHA3_SAVE_CONTEXT: u8 = 1 << 6;
+
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub enum HashAlgorithm {
+    Sha3_224,
+    Sha3_256,
+    Sha3_384,
+    Sha3_512,
```
chrysn, line 62: We could benchmark whether it makes any difference to just say `Sha3(Variant)` here, but we can also always change this later. (At this time, I'd say "do it later", but if further thought went into whether or not to do that, this would be a good place to comment that).

embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+            HashAlgorithm::Sha3_512 => Variant::Sha3_512,
+        }
+    }
+}
+
+impl embedded_cal::HashAlgorithm for HashAlgorithm {
+    fn len(&self) -> usize {
+        self.variant().output_len()
+    }
+
```
chrysn, line 80: The from_cose_algorithm is missing. That's hard to spot because they are in by their old names (SHAKE), but it seems that at least two are in there. (https://datatracker.ietf.org/doc/html/rfc9054#name-shake-algorithms has the definition that links to the FIPS spec).

embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+impl embedded_cal::HashProvider for super::Nrf54l15Cal {
+    type Algorithm = HashAlgorithm;
+    type State = HashState;
+    type Output = HashOutput;
+
+    fn init(&mut self, algorithm: Self::Algorithm) -> Self::State {
+        HashState {
+            variant: algorithm.variant(),
+            state: None,
+            buf: [0u8; MAX_RATE],
```
chrysn, line 166: The buffering does indicate that we might want to add a plumbing layer later -- but let's wait a bit, and see if maybe what we need is not so much a plumbing layer, but a set of tools ("reverse plumbing"? "handyman's toolbox?") that implements this concrete pattern from below or as a general utility function.

But agreeing with the PR description: The best time to do that is when a 2nd HW acceleration comes in.

embedded-cal-nrf54l15/src/sha3.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    fn finalize(&mut self, mut instance: Self::State) -> Self::Output {
+        let rate = instance.variant.rate();
+
+        // FIPS 202 multi-rate padding: place 0x06 after the last message byte, then
+        // OR 0x80 into the last byte of the rate block. When buf_len == rate - 1
+        // both writes land on the same byte, producing 0x86.
+        instance.buf[instance.buf_len] = 0x06;
+        instance.buf[rate - 1] |= 0x80;
```
chrysn, line 196: Is the buffer zeroed anywhere that I missed?

Otherwise the bytes between one-past-buffer-end and the last byte of the per-rate buffer would contain what was in the block before, that'd be odd.

We only have very short test vectors; maybe adding longer ones makes this apparent anyway.

embedded-cal-software-demo/src/hash.rs
```diff
@@ -196,15 +196,15 @@ impl<EC: ExtenderConfig> embedded_cal::HashAlgorithm for HashAlgorithm<EC> {
     fn from_ni_id(number: u8) -> Option<Self> {
         match number {
             1 => Self::from_cose_number(-16),
-            _ => None,
+            _ => HashAlgorithmOf::<EC::Base>::from_ni_id(number).map(HashAlgorithm::Direct),
```
chrysn, line 199: Please pick those changes into a commit of their own, describing what bug this fixes (I think it's "fix(software-demo/hash): let algorithm constructor fall through to base")

This will make the unavoidable merge conflicts trivial to solve.

chrysn (commented):
Mostly looks good, see concrete comments. In particular, please add nontrivial test vectors with lengths exceeding a first block.

chrysn:
The test vector situation is probably easier when we just take those from https://github.com/lake-rs/embedded-cal/pull/91 (which is approaching merging).

embedded-cal-nrf54l15/src/sha3.rs
WilliamTakeshi: I think SHAKE and SHA3 are different. re-reading the FIPS202 seems like SHAKE is a SHA-3 XOF.

by my understanding, SHA3 is always the same output len, while SHAKE can have different ones

WilliamTakeshi:
I will wait for #91 lands so I can start fixing this PR

PR #91, opened 2026-06-25
[libcrux] Add libcrux-iot sha3 implementation

robinhundt:
Stacked on #95 

This PR adds the SHA-3 implementation of libcrux-iot that uses [libcrux-secrets][ls] for lightweight secret independence checking to the libcrux `HashProvider`.

The [libcrux-secrets][ls] crate mainly works by restricting the operations that you can perform on a secret value, when compiled with the `check-secret-independence` feature enabled. The sha-3 implementation internally uses the [`U8`][u8] type alias that either resolves to the plain `u8` type, or to a `Secret<u8>` type that restricts the available operations when compiled with the `check-secret-independence` feature.
Thus a compilation with that cargo feature enabled constitutes a kind of proof that these secret values are only used in a secret independent way.

I've added a `check-secret-independence` feature to `embedded-cal-libcrux` that enables the corresponding features on libcrux-iot-sha3 and libcrux-secrets. When that feature is enabled and embedded-cal-libcrux (including libcrux-iot-sha3) type-checks, this essentially re-proves that the sha3 implementation is secret independent.

Because the `embedded-cal::HashProvider` trait operates on plain `u8` instead of the libcrux-secrets [`U8`][u8] type aliases, the libcrux hash provider implementation internally calls [`classify`][classify]/[`declassify`][declassify] at the boundary between the embedded cal API and the librux-iot-sha3 one.

To-do:
- [x] adding test vectors and tests for the sha3 hash functions
- [x] add explanatory comments for de-/classify operations and check-secret-independence feature
- [x] add sha3 to ALGORITHMS.md

[ls]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/
[u8]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/type.U8.html
[classify]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/trait.Classify.html
[declassify]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/trait.Declassify.html


robinhundt:
This is ready for review now. Note that it is stacked on #95, so if this reviewed while the other PR is not merged yet, it is best to filter the commits to those starting from [39ee4c2](https://github.com/lake-rs/embedded-cal/pull/91/commits/39ee4c21ffc7bf08a091a0a222f6d0184b1c85fe).

embedded-cal-libcrux/src/hash.rs
```diff
@@ -110,37 +173,69 @@ impl<EC: ExtenderConfig> embedded_cal::HashAlgorithm for HashAlgorithm<EC> {
...
     #[inline]
     fn from_ni_id(number: u8) -> Option<Self> {
         match number {
             1 => Self::from_cose_number(-16),
-            _ => None,
+            9 => Some(HashAlgorithm::Sha3_224),
+            10 => Some(HashAlgorithm::Sha3_256),
+            11 => Some(HashAlgorithm::Sha3_384),
+            12 => Some(HashAlgorithm::Sha3_512),
+            number => HashAlgorithmOf::<EC::Base>::from_ni_id(number).map(HashAlgorithm::Direct),
```
WilliamTakeshi: software now shadows hardware

because we resolve everything on the match, we will give preference for the software version.
e.g. `Extender<EC>::from_ni_id(9)` will give preference for the libcrux version.

it should work something more like `from_ni_id` from `-rustcrypto`

embedded-cal-libcrux/src/hash.rs
```diff
@@ -110,37 +173,69 @@ impl<EC: ExtenderConfig> embedded_cal::HashAlgorithm for HashAlgorithm<EC> {
...
     #[inline]
     fn from_ni_name(name: &str) -> Option<Self> {
         match name {
             "sha-256" => Self::from_cose_number(-16),
-            _ => None,
+            "sha3-224" => Some(HashAlgorithm::Sha3_224),
+            "sha3-256" => Some(HashAlgorithm::Sha3_256),
+            "sha3-384" => Some(HashAlgorithm::Sha3_384),
+            "sha3-512" => Some(HashAlgorithm::Sha3_512),
+            name => HashAlgorithmOf::<EC::Base>::from_ni_name(name).map(HashAlgorithm::Direct),
```
WilliamTakeshi: same comment as software now shadows hardware

embedded-cal-libcrux/src/hash.rs
```diff
@@ -110,37 +173,69 @@ impl<EC: ExtenderConfig> embedded_cal::HashAlgorithm for HashAlgorithm<EC> {
...
+/// [li]: https://github.com/celabshq/libcrux-iot
+/// [ls]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/
 pub enum HashResult<EC: ExtenderConfig> {
-    Sha256([u8; 32]),
     Direct(HashOutputOf<EC::Base>),
+    Sha256([u8; 32]),
+    Sha3_224([U8; SHA3_224_DIGEST_SIZE]),
+    Sha3_256([U8; SHA3_256_DIGEST_SIZE]),
+    Sha3_384([U8; SHA3_384_DIGEST_SIZE]),
+    Sha3_512([U8; SHA3_512_DIGEST_SIZE]),
```
WilliamTakeshi, line 236: this is just a question because I am curious

AFAIU, There is 2 ways of doing it.

1. The one right now, exposing `U8` on HashResult and adding `.declassify_ref()` in `as_ref`
2. having `[u8; SHA3_224_DIGEST_SIZE]` and usign `.declassify()` on finish

Whats the pros and cons of exposing the `U8`? Is it useful for creating more proofs on things on top? or makes no diff and its just preference?

.github/workflows/build-and-test.yml
```diff
@@ -25,6 +25,8 @@ jobs:
       run: cargo fmt --check
     - name: run `cargo check`
       run: RUSTFLAGS="-D warnings" cargo check --workspace
+    - name: Check embedded-cal-libcrux secret independence
+      run: cargo check -p embedded-cal-libcrux --features check-secret-independence
```
WilliamTakeshi: usually `RUSTFLAGS="-D warnings"` to also fail on warning is useful in cargo check, I never used this `check-secret-independence` featureflag, so I dont know if its a warning or an error

WilliamTakeshi (commented):
Really good, just have some questions about how secret-independence works, but everything else makes sense!

.github/workflows/build-and-test.yml
robinhundt: If there would be an issue with secret independence when the feature is enabled, it would result in a hard compile error. But still, it is good practice to deny warning, also under that feature, so I've added the rustflag.

embedded-cal-libcrux/src/hash.rs
robinhundt: Ohh good catch. I've changed it to be like in the rustcrypto provider.

embedded-cal-libcrux/src/hash.rs
robinhundt: Yes, those are the two ways and initially I was using the second one.

But then, based on the discussion with Christian about the assoc types and that we might be able to pull the libcrux-secrets types a bit higher, I wanted to see where the boundary is for exposing this currently. The `HashResult` type is as far as it goes, without making changes to the core embedded-cal traits.

Because the `HashResult` type is also made public in this PR, a user of the libcrux provider could use the the secrets based hash digest in that enum, e.g. for better compatibility with other libcrux APIs that use these types.

So there are some slight benefits to exposing it here, but mostly it was a matter of preference and trying to see where the boundary is.

WilliamTakeshi: Ok, its was just curiosity, makes sense exposing. Anything its really easy to change from one to another

WilliamTakeshi (approved):
Thanks for the hard work!

On my side, I think everything LGTM 🚀

embedded-cal-libcrux/src/lib.rs
```diff
@@ -9,6 +26,8 @@ use libcrux_sha2::Digest;
 mod aead;
 mod hash;
 
+pub use hash::{HashAlgorithm, HashResult, HashState};
```
chrysn: Exposing the types is a sensible thing to do. We never documented this in the architecture, but I think it's general practice around trait impls: Every implementation *may* provide additional traits or even the concrete type, and then users can use that directly.

But do you need all, including Algorithm and HashState? (If so, those would need non_exhaustive too.)

embedded-cal-libcrux/src/hash.rs
```diff
@@ -99,48 +158,95 @@ impl<EC: ExtenderConfig> Eq for HashAlgorithm<EC> {}
...
+///
+/// # Secret Independence Checking
+///
+/// The embedded-cal-libcrux  [`HashProvider`] uses a SHA-3 implementation
+/// provided by [libcrux-iot][li] that integrates with the [libcrux-secrets][ls] crate
+/// for lightweight secret-independence checking. This hash result enum exposes
+/// the SHA-3 digests as arrays of the [`tyalias@U8`] type alias from libcrux-secrets.
+///
+/// [li]: https://github.com/celabshq/libcrux-iot
+/// [ls]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/
```
chrysn, line 229:
```suggestion
/// [ls]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/
#[non_exhaustive]
```

As this is now pub-for-real and not just technically-pub-but-you-won't-find-it, we need to be a bit (and just a bit, at this stage) more careful about API stability.

testvectors/src/sha3.rs
```diff
@@ -0,0 +1,235 @@
...
...4210d50a3caa60b1b9db685f7704ae2d7302b18e8261052b779139747f462a6610a37252b170afbfce905fb6f7fb8c2b6100ee231507f403fee88ba5561580d4de4cdf600bf9e9816c9da1e1d2b91a1d966d04cdb98d3be55fb77af2daeeed750b8b60b494accaa12441d372afb3d47e7395b9e0e867595a1a6c8bff8638bcb138ddcac2f3efbf89762b68ebd77247c89929620f1a3cb8dcaf9632fde0996b33e6b2621da25924b4e2c8d6bff28ae0867786919ad763e6d79fc304a06277955795a7cb17186fb6bdfa98a16189544b228f3bcd3698737ff55b6185799459b796a63c6a61cea9d20f1e296d62f474c43750b77944e5f1c09072f019dbeeb64e9bc8dec4605d8e0322cdd97f56cc43084f5c983a584855654366fd5659ea23c6c15e1d7da51d82c683aa477b9f896563a5134c64e32814ea88b7f7af760f18bc91e656da92b72e98bc03f1c6bfb442830305529d681dc6bccae66da9b2e61b9c97e2397fdb92f7f6369b470529c570c2d3b329487981d148a462cdb992d792e34dd233e1c239657b8da0d59b804566cf81ad5f0a7a0ccb3a8fbda673887c153d2e56c484f9230d752be52c1e35bc9af5a7446237fc072afef777665c264c18e6a3c059fde2e8368f9bb898f1cc8393d1bf18b1757219670275f0bbc7deb0248c68af929111e19737479bcabab732d7e033aaeb277eac05e185e9e56b2450beaac784dd0308b7a5e8ca1f2fcd8852ddad9f7b7de264478e1891a391aa89964dae5ad0b7a829c2c9209db346ceb26c1b967cfac82ad574761443be3f0a910968239d23b11507ab978b3ce89e22b7d7283736b9786544ab4460f"
```
chrysn, line 52: I appreciate long test vectors, but that's a big blob -- does this come from any test vector set in particular (if so, please credit), or if not, is this just a random long string you generated and compared to other impls (also good, but please say so)?

embedded-cal-libcrux/src/lib.rs
```diff
@@ -1,6 +1,23 @@
...
+//! independent on the source code level.
+//!
+//! <div class="warning">
+//!
+//! The `check-secret-independence` feature is not intended to be enabled unconditionally.
+//! You should only enable it locally/in CI.
+//!
+//! </div>
+//!
+//! [ls]: https://docs.rs/libcrux-secrets/latest/libcrux_secrets/
```
chrysn, line 25: The pub types are added for this purpose, but I don't see any text on what to do with them; can you add a line here on how to use them? (But if it's complicated, adding this in a follow-up works too.)

chrysn (commented):
Almost ready, please see probably easy-to-addess comments.

embedded-cal-libcrux/src/lib.rs
robinhundt: True. `HashState` needs to be pub so that users have a choice of working with a `libcrux_secrets::U8` if they want, but the other two types are not necessary. I've removed them from the pub use.

chrysn: Seeing an example would help see how HashState is the relevant one.

My best assumption is that it's about sending in data (through `.update()`) that is already classified. If that is the case, I think the more idiomatic solution would be to have something like

```rust
impl<EC> Extender<EC> {
    pub fn update_with_classified(&mut self, instance: &mut Self::State, data: &[U8]) {
        match instance {
            HashState::Sha3_224(s) => s.0.update(data),
            …
            _ => panic!("attempted to use classified update with non-formally-verified back-end"), // or just unclassify and let the proof assistant show that we never really hit an unclassify case
        }
    }
}
```

because from the user's point of view it would almost not differ from the regular code path (that involves `.update()`).

chrysn: Then, we could do without pub associated types. (The `type DhProvider` of the top-level Cal is already Self, which is alreasdy pub)

testvectors/src/sha3.rs
robinhundt: The input values come from the `SHA256HASHES` constant in the testvectors crate. I computed the expected values using rust-crypto/sha3 version 0.12.0. I had that information in the commit message when I added the testvectors, but arguably that should have been in the source, as it also took me a bit now to find out how I derived them. I've added a doc comment to the sha3 explaining the provenance.

embedded-cal-libcrux/src/lib.rs
robinhundt: I've added a short sentence where I tried to briefly explain the relation of secret independence checking and the `HashResult` type.

embedded-cal-libcrux/src/lib.rs
chrysn: Thanks: my assumption was wrong, and the current branch makes things clearer.

We might still want to add a `.result_classified()` method at some point (again, pushing the workflow to be more similar than the usual workflow), but that'd be for later.

PR #92, opened 2026-06-25
noop: remove an unnecessary copy

WilliamTakeshi:


WilliamTakeshi:
I will just merge this one, since its a simple one that doesn't change anything in the architecture, just clean a dumb decision that I did prior.

PR #93, opened 2026-06-29
[libcrux-aes] fix missing copy in decrypt

robinhundt:
The current implementation of `decrypt_in_place` for the libcrux-aes
implementation is not actually in-place and allocates a temporary buffer.
The decryted message was not copied out from the buffer into the
initial message slice.

This PR adds a small sanity decrypt-encrypt test and adds the missing copy.

This probably slipped through as there were no tests for the libcrux-aes-gcm implementation. I've added a small one in this PR to validate the issue and the fix. Ideally, we'd also have test vectors for aes-gcm like for ccm, which can be reused across the implementations.

### DIFF
--- a/embedded-cal-libcrux/src/aead.rs
+++ b/embedded-cal-libcrux/src/aead.rs
@@ -214,6 +214,15 @@ impl<EC: ExtenderConfig> embedded_cal::AeadAlgorithm for AeadAlgorithm<EC> {
             AeadAlgorithm::AesGcm256 => libcrux_aesgcm::AesGcm256::NONCE_LEN,
         }
     }
+
+    fn from_cose_number(number: impl Into<i128>) -> Option<Self> {
+        let number = number.into();
+        match number {
+            1 => Some(AeadAlgorithm::AesGcm128),
+            3 => Some(AeadAlgorithm::AesGcm256),
+            _ => AeadAlgorithmOf::<EC::Base>::from_cose_number(number).map(AeadAlgorithm::Direct),
+        }
+    }
 }
 
 impl<EC: ExtenderConfig> Clone for AeadAlgorithm<EC> {
--- a/embedded-cal-libcrux/src/aead.rs
+++ b/embedded-cal-libcrux/src/aead.rs
@@ -274,10 +274,7 @@ impl<EC: ExtenderConfig> AsRef<[u8]> for Tag<EC> {
 
 #[cfg(test)]
 mod tests {
-    use embedded_cal::{AeadProvider, Cal};
-    use libcrux_aesgcm::NONCE_LEN;
-
-    use crate::{Extender, ExtenderConfig, aead::Key};
+    use crate::{Extender, ExtenderConfig};
 
     struct TestConfig;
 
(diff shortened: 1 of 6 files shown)


jschneider-bensch:
I'll separately update the APIs on the libcrux side to use `ExactSizeIterator` for the AAD and will remove the temp buffer and final copy here in a follow-up.

embedded-cal-libcrux/src/aead.rs
WilliamTakeshi, line 157: On alg::decrypt, is the order right? on `libcrux-traits` I see

```rust
    fn decrypt(
        plaintext: &mut [U8],
        key: &Key<Self>,
        nonce: &Nonce<Self>,
        aad: &[u8],
        ciphertext: &[u8],
        tag: &Tag<Self>,
    ) -> Result<(), DecryptError>;
```

just wanna make sure the plaintext/ciphertext are not swapped

WilliamTakeshi (commented):
Hey @robinhundt , thank you so much.

On tests, can you add like testvector/src/lib.rs? then I could easily use the same testvectors for the hw implementation. It also needs a check if the ciphertext/tag are the ones expected.

Right now we are just testing if the message is the same as the start, but if we add noop implementations for encrypt/decrypt, it would pass the tests.

embedded-cal-libcrux/src/aead.rs
jschneider-bensch: In the in-place API on top of this, `message` holds the ciphertext before decryption and should hold the decrypted message afterwards. Because with the `libcrux-traits` APIs used here (until we change that in a follow-up) that do not provide in-place decryption, a temporary buffer `ciphertext` is created into which libcrux will write the message decrypted from `message`, and which will be copied back into `message` afterwards. So I think the ordering is correct, but the naming is confusing, since `ciphertext` will only ever hold decrypted plaintext.

WilliamTakeshi: Thanks for the clarification. I tested locally and you are right!

robinhundt:
Hi @WilliamTakeshi, you mean adding test vectors like for AES-CCM? I see that for AES-CCM you have a small number of test vectors from RFCs/NIST SP 800-38C. For AES-GCM, I'm not aware of any authorative and small number of vectors like for CCM.

There are the vectors from [wycheproof](https://github.com/C2SP/wycheproof) and [NIST CAVP](https://csrc.nist.gov/projects/cryptographic-algorithm-validation-program/cavp-testing-block-cipher-modes), but these are larger files (would increase the repo's size) and would require parsing code.  

In the future, I think it makes sense to use these larger known answer test files, but I'm not sure if that should be included in this PR. For this PR, I could generate a few test vectors using a different implementation (or select some from wycheproof) and include them in the testvectors source files like for CCM.

robinhundt:
@WilliamTakeshi I've added test vectors from wycheproof to the test-vectors crate (which I've split up a bit more into modules) and now use these for the libcrux aes-gcm integration.

If you rather want these testing changes in a separate PR I can split it up.

robinhundt:
I've added the missing SPDX headers. One thing I'm unsure about is how to license the aes-gcm testvectors file. Wycheproof itself is [only licensed as Apache-2.0](https://github.com/C2SP/wycheproof/blob/main/LICENSE), so that is what I added for that file

WilliamTakeshi:
Thanks! LGTM :rocket: 

But before merging, @chrysn  do you know how license works on this case? I also don't know how this part works

WilliamTakeshi (approved):

WilliamTakeshi:
Merged since (by my understanding), the apache2 is only on the testvectors and we can rollback if needed in the future. Also testvectors are dev-dependency only, so it shouldn't change the license of the library

PR #94, opened 2026-07-01
[experiment] Aad on nrf54 

WilliamTakeshi:
This PR remove the copy the AAD on a buffer.

Before

CCM encrypt/decrypt built the whole authenticated header in one [u8; 288] stack buffer: it copied b0, the 2-byte AAD length prefix, and then every chunk of the AadGenerator into that buffer, and finally handed the buffer as a single  descriptor to the nRF54 scatter-gather engine.

After

We no longer copy the AAD. The header is fed as a sequence of descriptors that point directly at the caller's memory:

1. push(b0) — realign
2. push_raw(aad_len_prefix) — the 2-byte length, RAW (sub-word)
3. push_raw(chunk) for each AadGenerator chunk, streamed straight from the caller
4. a final realigning descriptor: the zero-pad (with DMATAG_IGN) when the header isn't block-aligned, otherwise the last AAD chunk itself

### DIFF
--- a/embedded-cal-nrf54l15/src/aead.rs
+++ b/embedded-cal-nrf54l15/src/aead.rs
@@ -33,6 +33,46 @@ const AES_CMD_CCM_DECRYPT: u32 = AES_CCM_MODE | 1; // 0x2001
 
 use crate::descriptor::{DescriptorChain, Input, Output, dmatag_ign};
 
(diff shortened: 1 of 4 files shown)


chrysn:
Good thing: This looks working.
Bad thing: It means that we *do* need to generic-bound the iterator.

PR #95, opened 2026-07-07
[libcrux-aes] Update APIs and add AES-CCM to supported algorithms

jschneider-bensch:
This PR

- updates the APIs to use the in-place encryption and decryption, as well as streaming AAD APIs for AES-GCM
- adds AES-CCM-128 and AES-CCM-256, along with short tag variants for each (8 bytes of authentication tag instead of 16).

Please let me know if you'd rather get those as separate PRs.

### DIFF
--- a/embedded-cal-libcrux/src/aead.rs
+++ b/embedded-cal-libcrux/src/aead.rs
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-License-Identifier: AGPL-3.0-only
 // SPDX-FileCopyrightText: Inria-AIO, Cryspen, and Christian Amsüss
 
 use libcrux_iot_aes::AeadConsts as _;
--- a/embedded-cal-libcrux/src/aead.rs
+++ b/embedded-cal-libcrux/src/aead.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: MIT OR Apache-2.0
 // SPDX-FileCopyrightText: Inria-AIO, Cryspen, and Christian Amsüss
 
-use libcrux_aesgcm::AeadConsts as _;
+use libcrux_iot_aes::AeadConsts as _;
 use libcrux_traits::aead::typed_owned;
 
 use embedded_cal::AeadProvider;
@@ -20,14 +20,14 @@ pub enum AeadAlgorithm<EC: ExtenderConfig> {
 
 pub enum Key<EC: ExtenderConfig> {
     Direct(AeadKeyOf<EC::Base>),
-    AesGcm128(libcrux_aesgcm::AesGcm128Key),
-    AesGcm256(libcrux_aesgcm::AesGcm256Key),
+    AesGcm128(libcrux_iot_aes::AesGcm128Key),
+    AesGcm256(libcrux_iot_aes::AesGcm256Key),
 }
 
 pub enum Tag<EC: ExtenderConfig> {
     Direct(AeadTagOf<EC::Base>),
-    AesGcm128(libcrux_aesgcm::AesGcm128Tag),
-    AesGcm256(libcrux_aesgcm::AesGcm256Tag),
+    AesGcm128(libcrux_iot_aes::AesGcm128Tag),
+    AesGcm256(libcrux_iot_aes::AesGcm256Tag),
 }
 
 impl<EC: ExtenderConfig> AeadProvider for Extender<EC> {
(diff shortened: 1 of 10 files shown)


jschneider-bensch:
Shouldn't get merged just yet, but is ready for review.

embedded-cal-libcrux/src/aead.rs
```diff
@@ -1,33 +1,75 @@
-// SPDX-License-Identifier: MIT OR Apache-2.0
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrysn: Unless this is copied out from AGPL'd code by someone else than the authors (who can just make their contributions here MIT-or-Apache), this doesn't need to have AGPL in here, or in the added licenses.

We can -- temporarily or permanently, although the latter would make some downstreams unhappy -- have AGPL dependencies, and that is completely orthogonal to our license: Just using the APIs doesn't make our *code* AGPL. (The resulting binary may be, I'd need to read up on AGPL, but that'll be visible to the user who compiles that in through the dependency chain, and we don't need to make our code AGPL).

embedded-cal-libcrux/src/aead.rs
```diff
@@ -74,6 +86,18 @@ impl<EC: ExtenderConfig> AeadProvider for Extender<EC> {
...
+            }
+            AeadAlgorithm::AesCcm128Short => {
+                Key::AesCcm128Short(<[u8; _]>::try_from(key).expect("key length mismatch"))
+            }
+            AeadAlgorithm::AesCcm256 => {
+                Key::AesCcm256(<[u8; _]>::try_from(key).expect("key length mismatch"))
+            }
+            AeadAlgorithm::AesCcm256Short => {
+                Key::AesCcm256Short(<[u8; _]>::try_from(key).expect("key length mismatch"))
+            }
```
chrysn: Minor nit: Moving the `.expect()` at the end of the function will make code smaller (because there are fewer panic locations).

chrysn (commented):
LGTM, except where noted explicitly.

As a small and procedural remark: I find `git absorb` very useful to, when later spotting small changes, to make those into fixup commits that can be autosquashed before merging. (It's not great for refactoring due to how it splits things up when moving lines around, that's better explicitly `--fixup`'d, but very useful for typo commits).

embedded-cal-libcrux/src/aead.rs
```diff
@@ -125,101 +164,88 @@ impl<EC: ExtenderConfig> AeadProvider for Extender<EC> {
...
         match self {
             AeadAlgorithm::Direct(a) => a.nonce_length(),
-            AeadAlgorithm::AesGcm128 => libcrux_aesgcm::AesGcm128::NONCE_LEN,
-            AeadAlgorithm::AesGcm256 => libcrux_aesgcm::AesGcm256::NONCE_LEN,
+            AeadAlgorithm::AesCcm128
+            | AeadAlgorithm::AesCcm128Short
+            | AeadAlgorithm::AesCcm256
+            | AeadAlgorithm::AesCcm256Short
+            | AeadAlgorithm::AesGcm128
+            | AeadAlgorithm::AesGcm256 => libcrux_iot_aes::NONCE_LEN,
```
WilliamTakeshi, line 224: NOUNCE_LEN = 12 on `libcrux_iot_aes`.

This dont match with https://datatracker.ietf.org/doc/html/rfc9053#name-aes-ccm

WilliamTakeshi (commented):
Looks good, I was just wondering about the NONCE_LEN when reading the libcrux-iot-aes code

embedded-cal-libcrux/src/aead.rs
jschneider-bensch: That is a good catch! Libcrux actually implements the AES-CCM configurations as specified for use in TLS in [RFC6655](https://www.rfc-editor.org/info/rfc6655), which specifies a 12-byte nonce.
I will update the algorithm descriptions here.

If you would rather we provide the COSE configuration now I could replace the TLS configuration with the COSE configuration in libcrux-iot and we'll work on supporting both in the future. Otherwise we'll implement the COSE configuration at a later point.

embedded-cal-libcrux/src/aead.rs
jschneider-bensch: I've dropped the AGLP licensing commit here, since we are in the process of relicensing libcrux-iot in any case.

embedded-cal-libcrux/src/aead.rs
jschneider-bensch: Not sure if this is what you meant, but I've pulled the converstion out into a helper function that should only be generated for the two key sizes.

chrysn: I thought about `match { ... }.expect("key length mismatch")`. But this also achieves that deduplication, thanks.

embedded-cal-libcrux/src/aead.rs
chrysn: I couldn't check yet if you have changed this already or not, but either way, mostly affecting the naming: While our current algorithm constructors mostly do that from COSE, embedded-cal should also stay suitable for TLS, allowing construction from TLS numbers. So if you already have AES-GCM-as-specified-for-TLS, that can stay in, and I can add a suitable constructor to the traits as part of the enumeration.

WilliamTakeshi (approved):
LGTM! 🚀

jschneider-bensch:
FYI, as the AES implementation in `libcrux-iot` is now merged into our `embedded-cal` branch, I've updated the git dependency here to refer to that branch. This will be the branch we'll do an alpha release from eventually as well.

chrysn (approved):
Thanks, LGTM.

As far as I am concerned, we can already merge this without waiting for the -alpha release: We're already juggling enough branches, and having a git patch dependency is not harmful at this stage (esp. with the promise that there will be a -alpha uploaded to crates.io).

chrysn:
I see that 10a65cfa9175e8f78198f24ba450c17090622945 is already merged into main; testing the direct update…

chrysn:
Nope: some patch is still needed because libcrux-iot components are used with non-main branch patches, so a plain "remove the patch/pinning and update" won't work.

Nonetheless, I'd merge this now to make progress on other PRs easier.

PR #102, opened 2026-08-06
Add HMAC and HKDF implementations for rustcrypto

WilliamTakeshi:
Working on lakers, I saw that we were still missing some implementations for rustcrypto

This should finish all software implementation on this side (as of today)

### DIFF
--- a/ALGORITHMS.md
+++ b/ALGORITHMS.md
@@ -47,9 +47,12 @@ Limitation in AAD streaming or message size are subject to ongoing work.
 |-----------|----------------|-------|
 | HMAC w/ SHA-256 | rustcrypto | |
 | HMAC w/ SHA-256 | software-demo | |
+| HMAC w/ SHA-256 | stm32wba55 | |
 
 # HKDF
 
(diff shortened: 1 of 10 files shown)


chrysn (approved):
Thanks, LGTM.

I've taken the liberty to push a note in ALGORITHMS that neither of the software impls uses the underlying SHA-256 acceleration (if any). For the softwaredemo, I have uncommitted code sitting here for longer than it should have sat; for the rustcrypto, that'd require impl'ing their trait on the underlying type or a wrapper thereof, which should be possible, but is not crucial to have.

(Thus I won't merge this without your eyes on that change; if it's good, please merge).

WilliamTakeshi:
rebased and fixed clippy

chrysn:
This now also practically pulled in #106 -- I'd guess due to a rebase onto that branch rather than onto main.

Please be careful with that kind of operations and immediate merges -- sadly, GitHub is not quite helpful there (a good forge should, for example, indicate whether or not a force push altered the tree, which a plain squash would not).

(But to fix it, let's continue discussion there).

PR #106, opened 2026-08-19
adopt a llm usage policy

WilliamTakeshi:
like discussed in [matrix](https://matrix.to/#/!HKnrmXbIiYsaCwImIV:matrix.org/$F9GkgJkJruC82-GhEzI4l9TA_pRXiQV5yGHSRWIZHrA?via=matrix.org&via=mozilla.org&via=academiccloud.de), we could adopt an llm usage policy.

`rust-lang/rust` adopted one that I think its a good compromise.

In this PR I added a policy heavily based on their policy, just removing parts that doesn't concern us (their tooling, teams, etc.)

Everyone, feel free to add/remove/edit/comment anything you think its necessary

### DIFF
--- a/policies/llm-usage.md
+++ b/policies/llm-usage.md
@@ -1,7 +1,6 @@
 # LLM Usage Policy
 
 This is a moderation policy for how LLMs are used in `lake-rs/embedded-cal`.
-For additional information about the policy itself, see [the appendix](#appendix).
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


chrysn (commented):
This basically reflects what we discussed before the PR was created, I agree with the content.

Just one note before merging:
* The PR description states that this is based on Rust's policies -- then, the text should reflect that (in a comment, and if it takes text verbatim, also in the copyright statement that so far claims it's ours alone).
* I think it should be a bit more explicit in that this might change over time, in particular when more standard text becomes available.

Things for after merging:
* I have yet to document our processes; integrating that may move this document.

chrysn:
This got merged accidentally through #102.

To fix this, please apply fixes to commit 7ca3d7e79bf6b565fe1ea5c25fbef166078a91a1, and force-push that branch into this PR.

WilliamTakeshi:
> This got merged accidentally through https://github.com/lake-rs/embedded-cal/pull/102.

Sorry!!! I rebased against a dirty main on my computer...

I reverted on main and re-did all 4 commits here.

chrysn:
Reverting works too, so we can continue where we left off:

> Just one note before merging:
>
> -    The PR description states that this is based on Rust's policies -- then, the text should reflect that (in a comment, and if it takes text verbatim, also in the copyright statement that so far claims it's ours alone).
> -    I think it should be a bit more explicit in that this might change over time, in particular when more standard text becomes available.

PR #110, opened 2026-09-09
add tests to test Ec trait directly

WilliamTakeshi:
Today the trait is tested using the DhProvider, but whentestsing hardware it is just easier having the test directly on EC.


chrysn:
Labelling MVP-critical for its test coverage, although I hope we can get it in with the fixes of #111.

PR #111, opened 2026-09-09
Experiments on #110

chrysn:
I think that #110 could become a -software-demo implementation of what is being tested here, without a lot of extra effort -- and then rather than have complicated tests that tap directly into the back-end, we use the existing tests on the software-demo on the back-end.

But first, I'll have to understand why some things are there.


chrysn:
Hmpf, not getting this far enough before going off to bed, but @WilliamTakeshi, does this make sense to you so far?

WilliamTakeshi:
Just checking if I understand it.

Instead of writing the test for each one like I did, we change that `-software-demo` becomes a consumer of the EC plumbing and test on this level?

I like the idea, I think its elegant and should simplify to test it

chrysn:
Yes.

Ideally, this will also make the `impl Dh for Nrf` and `impl Dh for Stm` moot; those have been slated for unification already when the EC plumbing came in.

chrysn:
Rolled back now to the parts that are complete -- they don't quite fix all of what I think we could do better than #110, but it fixes things that led to #110 being what it is (namely, software-demo not doing a good job in illustrating the separation of plumbing and high-level roles).

I'll do a round of fixing of merge conflicts before marking this as ready-for-review; the "move functionality of the tests into software-demo" step has become PR #127.

embedded-cal-nrf54l15/tests/integration.rs
```diff
@@ -27,66 +43,86 @@ mod tests {
...
         for v in testvectors::dh::RFC7748_X448 {
-            v.test_with(state.cal.dh());
+            v.test_with(state.cal.as_mut().unwrap().dh());
+        }
+    }
+
+    #[test]
+    fn test_ec_plumbing_p256(state: &mut super::TestState) {
+        for v in testvectors::dh::RFC5903_P256 {
+            v.test_plumbing_x25519(state.cal.as_mut().unwrap());
```
WilliamTakeshi: should be test_plumbing_p256

embedded-cal-software-demo/src/hash.rs
```diff
@@ -186,25 +183,34 @@ impl<EC: ExtenderConfig> embedded_cal::HashAlgorithm for HashAlgorithm<EC> {
...
         let number: i128 = number.into();
 
+        // Sanity check: Having the software implementation on top of an implementation without
+        // SHA2 does not make sense.
+        const {
+            assert!(
+                EC::IMPLEMENT_SHA2SHORT
+                    >= <EC::Base as Sha2Short>::SUPPORTED | EC::IMPLEMENT_SHA2SHORT_PLUMBING
+            )
+        };
```
WilliamTakeshi, line 193: is this right?

for example:

`IMPLEMENT_SHA2SHORT=true, SUPPORTED=false, PLUMBING=false` evals to `true`. but should  not pass the sanity check.

I think `>=` should be `<=`

WilliamTakeshi (approved):
LGTM, just one typo and one comment on the sanity check that I am not 100% sure!

Also

Closes: #110

chrysn:
Argh, fatfingered a Close: Thanks, both were mistakes, and are now fixed.

PR #114, opened 2026-09-10
[libcrux-dh] add p256 dh provider

robinhundt:
stacked on #91 

This PR adds a libcrux backed DH provider that either uses a base DH provider or a libcrux-iot [P256 implementation](https://github.com/celabshq/libcrux-iot/tree/embedded-cal/libcrux-iot/p256).

I have started with the DH provider as that required less changes to the p256 crate. The next step would be to expose the relevant parts of libcrux P256 and add an `EcPrimitives` implementation.


embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+        alg: Self::Algorithm,
+        data: &[u8],
+    ) -> Result<Self::PublicKey, embedded_cal::ImportError> {
+        match alg {
+            DhAlgorithm::P256 => {
+                if !validate_public_key(data) {
+                    return Err(embedded_cal::ImportError);
+                }
+                let pk = data.try_into().map_err(|_| embedded_cal::ImportError)?;
+                Ok(PublicKey::P256(P256PublicKey(pk)))
```
WilliamTakeshi, line 155: `validate_public_key` panics when data is not `.len() == 64`. I think if you check pk before then do the validate it works.

Should this be a bug in libcrux-iot-p256?

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+                .map(VisibleSecretKey::Direct),
+        }
+    }
+
+    fn export_publickey_bytes<'p>(
+        &mut self,
+        public: &'p Self::PublicKey,
+    ) -> impl AsRef<[u8]> + use<'p, EC> {
+        match public {
+            PublicKey::P256(pk_bytes) => Either::Own(pk_bytes),
```
WilliamTakeshi, line 137: from the trait: `export_publickey_bytes`
> ... this is defined as the compact representation

I think here we are just sending all bytes `x||y`, I think best keep the trait consistent

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+            PublicKey::Direct(pk_bytes) => {
+                Either::Direct(self.0.dh().export_publickey_bytes(pk_bytes))
+            }
+        }
+    }
+
+    fn import_publickey_bytes(
```
WilliamTakeshi, line 144: also, same as export_publickey_bytes, should be the compact representation

on other impls we usually use p256_recover_y to recover the y part

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum DhAlgorithm<BA> {
+    P256,
+    Direct(BA),
+}
+
+impl<Base: embedded_cal::DhAlgorithm> embedded_cal::DhAlgorithm for DhAlgorithm<Base> {
+    fn output_length(&self) -> usize {
+        SECRET_LEN
+    }
```
WilliamTakeshi, line 30: something more like `-rust-crypto`

```rust
    fn output_length(&self) -> usize {
        match self {
            DhAlgorithm::P256 => 32,
            DhAlgorithm::Direct(d) => d.output_length(),
        }
    }
```
so when we do for example `DhAlgorithm::X448`, it rightly chooses the one in the hardware implementation

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+            VisibleSecretKey::Direct(secret_key) => SecretKey::Direct(secret_key.into()),
+        }
+    }
+}
+
+impl AsRef<[u8]> for P256SecretKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.declassify_ref()
+    }
+}
```
WilliamTakeshi, line 229: does this contradict the comment 
```rust
// We don't want to provide access to the key bytes directly, so create this
// new-type that can be put into the pub enum
```

I think deleting this `impl` and putting `.declassify.ref()` in every use. but I am not sure

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,258 @@
...
+    ecdh_api::{EcdhOwned, PUBLIC_LEN, SECRET_LEN},
+    validate_public_key,
+};
+use libcrux_secrets::{ClassifyRef, DeclassifyRef, U8};
+use rand_core::Rng;
+
+use crate::{Extender, ExtenderConfig};
+
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum DhAlgorithm<BA> {
```
WilliamTakeshi, line 22: there is many things here that doesn't need to be pub, I think DhAlgorithm, VisibleSecretKey, SecretKey, etc. can be private. Makes it easier to read and adding the pub isn't a breaking change, while the opposite is.

WilliamTakeshi (commented):
Thanks! I am really happy to see this taking a good shape 😃 , added some comments and I think we also can start adding `#[derive(Zeroize, ZeroizeOnDrop)]` on some of these structs.

embedded-cal-libcrux/src/dh.rs
robinhundt: Oh, thanks for catching that. The `validate_public_key` is sadly only exposed under the `expose-hacl` feature which directly exposes the hacl extracted code that is not quite idiomatic Rust.

The panicking behavior is a documentation gap on the `validate_public_key` function and ideally, this function is also exposed in the higher level interface with a non-panicking API. But for this integration into embedded-cal, using the validate function is not necessary anymore with the change to the compact representation.

embedded-cal-libcrux/src/dh.rs
robinhundt: True, I should have looked more properly at the requirements of the trait. I've changed import/export to use the compact representation.

embedded-cal-libcrux/src/dh.rs
robinhundt: I'm now using the same sec1 compressed point to string conversion with sign of y hard-coded to 0 that is used in the rust-crypto provider and described [here](https://datatracker.ietf.org/doc/html/rfc9528#name-compact-representation).

embedded-cal-libcrux/src/dh.rs
robinhundt: Fixed.

embedded-cal-libcrux/src/dh.rs
robinhundt: The comment was a bit misleading. What I intended was to forbid "mutable access" so that they key validity can be upheld by construction.
But there is no reason to have these AsRef impls. A user should go through the export methods on the `DhProvider` trait, so I removed them.

embedded-cal-libcrux/src/dh.rs
robinhundt: They can't be private, because they're named in the public interface `DhAlgorithm`. E.g., changing `DhAlgorithm` to private results in:

```rust
error[E0446]: private type `dh::DhAlgorithm<<<<EC as ExtenderConfig>::Base as embedded_cal::Cal>::DhProvider as DhProvider>::Algorithm>` in public interface
  --> embedded-cal-libcrux/src/dh.rs:83:5
   |
24 | enum DhAlgorithm<BA> {
   | -------------------- `dh::DhAlgorithm<<<<EC as ExtenderConfig>::Base as embedded_cal::Cal>::DhProvider as DhProvider>::Algorithm>` declared as private
...
83 |     type Algorithm = DhAlgorithm<DhAlgorithmOf<EC::Base>>;
   |     ^^^^^^^^^^^^^^ can't leak private type

```

Note that these types are marked pub, but not nameable via the public API as they're in a private module and not public re-exported. I followed this convention from other providers I saw.
Having these types pub but not nameable makes them a bit more annoying for a user, e.g. they can't just be stored in a struct, but would need to be named via the `DhProvider` associated type. On the other hand, a rename of a type like `DhAlgorithm` would not be a breaking change.

I don't have a strong preference whether these types should be re-exported as pub or stay as pub in private types.

embedded-cal-libcrux/src/rand.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::{Extender, ExtenderConfig};
+
+// FIXME: This implementation is based on the one in embedded-cal-rustcrypto and intended for
+//  testing purposes for now.
+/// An implementation based on `getrandom`.
+///
+// FIXME: We should probably have some fast CSPRNG in self that is just seeded from getrandom.
+impl<EC: ExtenderConfig> rand_core::TryCryptoRng for Extender<EC> {}
+
+impl<EC: ExtenderConfig> rand_core::TryRng for Extender<EC> {
```
chrysn, line 13: I don't think libcrux should itself pick getrandom here. Instead, it should merely require that Self: TryRng, and forward the TryRng to the base.

That also saves this crate the discussion of any "should we go through the underlying RNG or a CSPRNG seeded from it", because that decision can be taken by the user depending on the system. (On RTOSes, there might be a global CSPRNG, and shipping an own would not make sense, whereas on desktop OSes, going through a getrandom syscall would easily slow things down).

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,275 @@
...
+    P256(P256SecretKey),
+    Direct(BSK),
+}
+
+// Also don't expose the public key bytes directly
+pub struct P256PublicKey([u8; PUBLIC_LEN]);
+
+impl P256PublicKey {
+    fn x(&self) -> &[u8; 32] {
+        self.0[..32].try_into().expect("slice has len 32")
```
chrysn: This will panic differently depending on whether it is too long or too short; this won't:


```suggestion
        self.0.try_into().expect("slice has len 32")
```

testvectors/src/dh.rs
```diff
@@ -106,3 +134,19 @@ pub const RFC5903_P256: &[EccVector] = &[EccVector {
     bob_public: &hex!("D12DFB52 89C8D4F8 1208B702 70398C34 2296970A 0BCCB74C 736FC755 4494BF63"),
     shared_secret: &hex!("D6840F6B 42F6EDAF D13116E0 E1256520 2FEF8E9E CE7DCE03 812464D0 4B9442DE"),
 }];
+
+/// [RFC5903][rfc] P256 test vectors with `pk = x||y`.
+///
+/// [rfc]: https://www.rfc-editor.org/info/rfc5903/#section-8.1
+pub const RFC5903_LARGE_PK_P256: &[EccVector] = &[EccVector {
```
chrysn: I think that this form has become unused in 7696bc97fc8927f0fbcce1561630bb98b4d5ced7 -- can we remove it?

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,275 @@
...
+                // Here, we reuse the libcrux p256 sec1 compressed decoding and just set the y sign to always
+                // be even.
+                // This trick is described in https://datatracker.ietf.org/doc/html/rfc9528#name-compact-representation
+                // which defines the compact representation.
+                if data.len() != 32 {
+                    return Err(embedded_cal::ImportError);
+                }
+                // Sec1 compressed repr first contains an octet designating the sign of y, even = 0x02 or odd = 0x03
+                // Then the 32 bytes of the x point, 33 bytes in total.
+                debug_assert_eq!(SECRET_LEN, 32);
```
chrysn:
```suggestion
                const { assert!(SECRET_LEN  == 32) };
```

(Preferring build-time failures over conditional run-time failures)

chrysn (commented):
Not through reviewing (next up is the bulk implementation), I'll come back to it, but some comments already.

embedded-cal-libcrux/src/dh.rs
robinhundt: `self.0` has a length of 64-bytes as libcrux-iot expects the public key bytes to be `x||y`. So the indexing is necessary here and both can never fail. I'll add a `get().expect()` to document why this slicing is safe here.

embedded-cal-libcrux/src/rand.rs
robinhundt: Mh, that seems a bit tricky to do. What I'm trying is to require `CryptoRng` on the `Base`, i.e.,
```rust
pub trait ExtenderConfig {
    type Base: Cal + Plumbing + rand_core::CryptoRng;
}
```
`TryCryptoRng` wouldn't work here, because `DhProvider::generate_visible` is infallible, so we need infallible randomness generation there.

But if I require `CryptoRng` on the base, I can't use `EmptyCal` as the base in the tests, as it doesn't implement that.

Would you be okay with an optional, intended for testing, getrandom-based impl of `CryptoRng` for `EmptyCal` similar to what I had here/is in the rust-crypto provider? This would add an optional dependency on getrandom to embedded-cal though, which I'm not sure about.

robinhundt: Mh, maybe some adapter like `WithRng<Cal, Rng>(pub Cal, pub Rng)` that delegates to a a `Cal` but also delegates to some Rng could help. Then for testing, this can be `type Base = WithRng<EmptyCal, TestRng>` and with an actual hardware base it would use the hardware rng.

I can try this out in `embedded-cal-libcrux` for now, but if you want, we can also move this to `embedded-cal`, as it is a generic adapter that could be useful for other providers.

embedded-cal-libcrux/src/dh.rs
chrysn: Thanks. I wish Rust had a built-in idiom for "This is a long enough array that I can get that constant range out of it reliably and it's an array again", but until then, this will do.

embedded-cal-libcrux/src/rand.rs
chrysn: It wouldn't need to fan out (and can't, because if it is the underlying Cal that impleents Rng, that can't be in there twice).

And yes it is not libcrux specific: in the end, a software-only solution would be eg. a `e_c_libcrux::Extender<e_c_getrandom::Extender<Empty>>`.

As we don't have that in place, I'd be fine with leaving this in, as long as we keep track of it in follow-up issues.

embedded-cal-libcrux/src/dh.rs
robinhundt: Yep, I wish that too. I had a look at the tracking issue of that feature https://github.com/rust-lang/rust/issues/90091, which doesn't have a lot of activity, but it led me to this issue https://github.com/rust-lang/rust-clippy/issues/16396 .

I wasn't aware of these `*_chunk` methods and have changed the code here to use them. I prefer that over the indexing/try_into dance.

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn shared_secret(
+        &mut self,
+        private: &Self::SecretKey,
+        public: &Self::PublicKey,
+    ) -> Result<Self::SharedSecret, embedded_cal::IncompatibleKeys> {
+        match (private, public) {
+            (SecretKey::P256(secret), PublicKey::P256(public)) => {
+                // FIXME: This feels somewhat like an abuse of the IncompatibleKeys error. P256::derive_ecdh should only
+                //  return an error if the keys are invalid, which should be impossible by construction. But the alternative
+                // is using an unreachable! here, which would crash the process if that "should be impossible" is wrong.
```
chrysn, line 204: I think an unreachable would be fine here, because that is a sensible result for data structure invariants being violated. (But quodlibet ex falso, and IncompatibleKeys is a fine outcome as well).

chrysn (approved):
LGTM.

I'm not happy with having getrandom in here precisely, but if we open an issue about enhancing that (referecning https://github.com/lake-rs/embedded-cal/pull/114#discussion_r4004497731), we can merge this now and refactor later.

embedded-cal-libcrux/src/dh.rs
robinhundt: I've become more cautious about `unreachable!`. We recently had a had case where our hmac-drbg implementation had a `unreachable!("this is fine because...")` which was indeed not fine and was reachable.

chrysn: Yeah, that's part of why erring in some vaguely meaningful way is OK too.

However, I expect that given that this *should* be an infallible thing, many users will (if they made sure the algorithms line up) justifiedly unwrap the result, and then it's the same mess (but they'll have to think about whether they did something wrong, whereas an unreachable! here clearly tells them that yeah that's a bug in the e-c implementation).

embedded-cal-libcrux/src/rand.rs
robinhundt: Mh, I'm not sure I quite understand what you're proposing. I've added a `WithRng` adapter that tries out one option for resolving this. I can remove this commit if you prefer another solution so that we can get this merged.

robinhundt:
I've reverted the branch back to commit https://github.com/lake-rs/embedded-cal/pull/114/commits/6c934815642c7e3db6b4ff4d7fab6e392f15c434 which was the one you approved. I'll propose the `WithRng` changes in a follow-up PR.

testvectors/src/dh.rs
```diff
@@ -62,6 +41,55 @@ impl EccVector {
...
+
+        let private_visible = cal
+            .import_secretkey_bytes(alg.clone(), private_bytes)
+            .unwrap_or_else(|_| panic!("failed to load {name}'s secret key"));
+
+        assert_eq!(
+            cal.export_secretkey_bytes(&private_visible).as_ref(),
+            private_bytes,
+            "{name}'s secret key did not round-trip through import/export"
+        );
```
WilliamTakeshi:
```suggestion
        // We cannot test export(import(x)) == x, because some implementations clamp on import
        // (the nRF54L15 back-end pre-clamps X25519 and X448 scalars as in RFC7748's decodeScalar),
        // and that operation loses information.
        // Instead we are testing the idempotence export(import(export(import(x)))) == export(import(x)).
        //
        // The scope keeps `exported`'s borrow of `private_visible` from outliving the check;
        // the returned `impl AsRef<[u8]>` may have a destructor, so NLL cannot end it early.
        {
            let exported = cal.export_secretkey_bytes(&private_visible);
            let reimported = cal
                .import_secretkey_bytes(alg.clone(), exported.as_ref())
                .unwrap_or_else(|_| panic!("failed to re-import {name}'s exported secret key"));
            assert_eq!(
                cal.export_secretkey_bytes(&reimported).as_ref(),
                exported.as_ref(),
                "{name}'s secret key did not round-trip through export/import"
            );
        }
```

WilliamTakeshi: export(import(x)) == x is not true on X25519 and X448 (tested on hardware), because clamping lose some information.

WilliamTakeshi (approved):
LGTM, just the tests on `test_dh_x25519` and `test_dh_x448` is failing on hardware. I suggest a change that its working here.

Check if makes sense, then you can merge it :))

```
(8/9) running `test_dh_x25519`...
[ERROR] panicked at testvectors/src/dh.rs:67:9:
assertion `left == right` failed: Alice's secret key did not round-trip through import/export
  left: [112, 7, 109, 10, 115, 24, 165, 125, 60, 22, 193, 114, 81, 178, 102, 69, 223, 76, 47, 135, 235, 192, 153, 42, 177, 119, 251, 165, 29, 185, 44, 106]
 right: [119, 7, 109, 10, 115, 24, 165, 125, 60, 22, 193, 114, 81, 178, 102, 69, 223, 76, 47, 135, 235, 192, 153, 42, 177, 119, 251, 165, 29, 185, 44, 42] (panic_probe panic-probe-1.0.0/src/lib.rs:104)
Firmware exited unexpectedly: Exception

 running `test_dh_x448`...
[ERROR] panicked at testvectors/src/dh.rs:67:9:
assertion `left == right` failed: Alice's secret key did not round-trip through import/export
  left: [152, 143, 73, 37, 209, 81, 159, 87, 117, 207, 70, 176, 75, 88, 0, 212, 238, 158, 232, 186, 232, 188, 85, 101, 212, 152, 194, 141, 217, 201, 186, 245, 116, 169, 65, 151, 68, 137, 115, 145, 0, 99, 130, 166, 241, 39, 171, 29, 154, 194, 216, 192, 165, 152, 114, 235]
 right: [154, 143, 73, 37, 209, 81, 159, 87, 117, 207, 70, 176, 75, 88, 0, 212, 238, 158, 232, 186, 232, 188, 85, 101, 212, 152, 194, 141, 217, 201, 186, 245, 116, 169, 65, 151, 68, 137, 115, 145, 0, 99, 130, 166, 241, 39, 171, 29, 154, 194, 216, 192, 165, 152, 114, 107] (panic_probe panic-probe-1.0.0/src/lib.rs:104)
Firmware exited unexpectedly: Exception

````

robinhundt:
Thanks for running it on the hardware and catching the clamping issue.

Doing `export(import(export(import(x)))) == export(import(x))` is not ideal, as this will not catch certain bugs in the export implementation. For example, if export by mistake just returns a zero initialized slice, this will not be caught by test.

One option might be to match on the cose ID of the curve and do the clamping for X25519 and X448 in the `test_key_import_export` function. But I'm not sure if this is putting too much curve specific logic into the testing code.

chrysn:
Can't we just add a field to the tests vector that has the clamped secret key in it, and then require that the exported one is either the original or the clamped form? (Alternatively we could only put clamped values in the test vectors, but I do see value in testing with unclamped input).

robinhundt:
I've tried that approach here https://github.com/celabshq/embedded-cal/pull/1 but am not really happy with it.

From my side, this PR here would be good to merge. I think at the moment, such issues with potential faulty export implementations are unlikely enough that this doesn't need to block this PR.

But, as I wrote in the other PR:
> I wonder whether the contract on import/export should maybe be tightened, so that it is guaranteed that they're idempotent. This would require a change in either the nrf or rust-crypto backend for these curves. The non-idempotency could easily trip people up, as it did for me.

But this can also be a follow-up issue if makes sense to have that property.

chrysn:
Follow-up tracked in https://github.com/lake-rs/embedded-cal/issues/122

PR #115, opened 2026-09-11
Introduce contribution and maintenance policies

chrysn:
Building on #106 (which is close to merging; ideally, if that just gets one commit on top that addresses the comments, this won't even have to rebase) because that sets up the directory structure and is also cross-referenced.

This PR does not aim to set policies, but rather to document how we've been doing things (or how we generally aspire to be, such as "aim[ing] to provide a friendly and inclusive atmosphere in which the project can grow"). We should have text on that somewhere visible approaching a release.

Please review for whether we agree in how we see this project running, and whether it is written down correctly. Noteworthy points where I diverge from established practice:

* crates.io owners should be a balanced set. Right now it is only @malishav, but as we talked about me managing the releases: Can you add me to the owners?
* We haven't really released anything yet, so how we release is new here (but that's how I've been doing it recently).
* So far we just put the same copyright notice in everywhere, but that doesn't really go well with new people joining (for whom this is also written). I don't see harm in diverging here: If someone adds a new file, they just put their name in there, and as any of us enhances it, we add those names again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,8 @@ by [Inria-AIO](https://aio.inria.fr/),
 [Cryspen](https://cryspen.com/)
 and [Christian Amsüss (@chrysn)](https://christian.amsuess.com/).
 
-We welcome additional contributors;
-at the current stage, that can mainly happen in the interface designing, scope and requirements discover
-happening [in the issue tracker](https://github.com/lake-rs/embedded-cal/issues).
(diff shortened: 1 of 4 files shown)


malishav:
Just invited you as an owner on crates.io, thanks for the ping

WilliamTakeshi (approved):

PR #118, opened 2026-09-14
libcrux: add hash update_with_classified method

robinhundt:
@chrysn I saw your [suggestion](https://github.com/lake-rs/embedded-cal/pull/91/#discussion_r4004259284) with the `update_with_classified` method too late. I like this way of having optional integration with libcrux-secrets in this way, so I added the method to the extender.

I chose not to panic in case the underlying hash algorithm is not using libcrux-secrets itself. There is no harm here in calling declassify, and I prefer that over having a higher risk of a reachable panic.


chrysn:
If passing it on doesn't upset the formal tools, great.

Can we un-`pub use` the associated types in this?

robinhundt:
> Can we un-`pub use` the associated types in this?

Which assoc types do you mean? This only re-exports the `HashResult` which I would keep.

chrysn:
Right, it's only one.

I mixed up a bit the hash result (where it is not envisioned that you'd need to consult the &mut Cal to get its data out) and operations on secrets (where the extraction does take a &mut Cal), so there's no function of the same style as `update_with_classified()` that would give you a hash still-classified hash result.

But maybe it'd make sense as a method on the algorithm enum, like `as_classified()`, that spares a user who would otherwise go matching the enum the hassle of keeping up with other algorithms?

robinhundt:
> But maybe it'd make sense as a method on the algorithm enum, like `as_classified()`, that spares a user who would otherwise go matching the enum the hassle of keeping up with other algorithms?

Do you mean an `as_classified()` on `HashAlgorithm` or `HashResult`? I don't see what the former would do. For the latter, an `HashResult::as_classified() -> impl AsRef<[U8]>` could be useful.

chrysn:
On the latter, `HashResult.as_classified()`, as you described.

One thing this would enable us to do is to have the enum's variants private, keeping the implementation flexible once we start adhering to semver guarantees. (enums can't really be pub but have their members private; the hack around it is usually to have an internal enum wrapped in a single-member struct).

robinhundt:
I've added the `as_classified` method, thanks for the suggestion!

But I'm not sure whether the pub enum in private module but not re-exported actually achieved what we (I think) wanted it to do. While the name of the type enum can't be written, the assoc type can be bound to a type alias, and then a value of the enum can be matched, with the variants named through the type alias.

<details>
<summary> Example </summary>

```rust
pub trait HashProvider {
    type Result;

    fn hash(&self) -> Self::Result;
}

pub struct Hasher;

mod private {
    use crate::HashProvider;

    pub enum HashResult {
        Hash([u8; 32]),
    }

    impl HashProvider for super::Hasher {
        type Result = HashResult;

        fn hash(&self) -> Self::Result {
            HashResult::Hash([0; _])
        }
    }
}
```

In another crate, or in this case integration test:

```rust
use enum_vis_test::{HashProvider, Hasher};

#[test]
fn test_vis() {
    let hasher = Hasher;
    type HashResult = <Hasher as HashProvider>::Result;
    match hasher.hash() {
        HashResult::Hash(hash) => dbg!(hash),
    };
}
```

</details>

I've changed the pub enums to private inner enums within public structs to guard against this. I'm not sure how likely is that a user would actually use the API in this way, but technically, changing the enum variants would be a breaking change due to this interaction, even though they are not publicly exported.

WilliamTakeshi (approved):
LGTM!

PR #119, opened 2026-09-14
[libcrux] Add `WithRng` adapter

robinhundt:
stacked on #114 

Context: https://github.com/lake-rs/embedded-cal/pull/114#discussion_r4004497731

This removes the `getrandom`-based implementation of `TryRng` for the libcrux extender. Instead, it is required that the `Base` Cal implementes `CryptoRng` (infallible because `DhProvider::generate_visible` is infallible).

Because `EmptyCal` does itself not implement `TryRng`, I've added a `WithRng` adapter that composes a `Cal` and a `TryRng` into one struct that implements both.

For testing purposes, we can then use a `type Base = WithRng<EmptyCal, rand::StdRng>`. 

LLM Disclosure: The idea of an `WithRng` adapter was proposed by claude code (Opus 5). I liked that idea and have written all code and docs in this PR by myself.


WilliamTakeshi (approved):

embedded-cal-libcrux/src/dh.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    fn generate_visible(&mut self, alg: Self::Algorithm) -> Self::VisibleSecretKey {
+        match alg {
+            DhAlgorithm::P256 => {
+                let mut rand = [0; _];
+                // libcrux p256 does not perform rejection sampling or sampling with extra bits.
+                // It only checks whether the provided random value would be a valid key or not,
+                // so we implement a simple rejection sampling here.
+                loop {
+                    self.0.fill_bytes(&mut rand);
```
chrysn, line 102: An extender should pass on traits (tedious as that currently is in Rust, maybe we should provide macros for the traits); then this can stay `self.fill_bytes`.

chrysn (approved):
GitHub's view is not helpful in the current stage of merge conflicts and not updating to a reflected main branch, but I've looked at 559d54c28feb5b4c12e1771ea711e2bdac0aed81 (or its cherry-picked version or a hypothetical merge, all the same):

LGTM, with one comment.

I'd like to take it a bit further so that both rustcrypto and this can use the WithGetrandom extender, and have convenience constructors for it (which comes from writing introductory docs, and thus from building a smoother new-user experience).

As I can't merge it anyway (or push commits), I'll take this state and work from it to also add that; we can build on the base reviews in here to later merge that then after your or William's review.

embedded-cal-libcrux/src/dh.rs
chrysn: Clarification: Pass-through *is* implemented, this change was probably just made before pass-through was complete.

PR #120, opened 2026-09-16
Add tests for hardware-accelerated libcrux P256

jschneider-bensch:
This PR adds demonstration crates that use `libcrux-iot-p256` with hardware acceleration via embedded-cal traits on the following boards

- nRF54l15, backed by `embedded-cal-nrf54l15`
- STMWBA55, backed by `embedded-cal-stmwba55`


chrysn (commented):
That's a good direction. IIUC we're more preparing this to be ready as a demonstrator than for immediate merging, so while working on resolving things such as the allocator, here's how I think this should be extended:

* The pieces from LibcruxTestVector can be used to move into an Extender. (I don't know how libcrux and libcrux-iot are related, so this might be standalone or move into e-c-libcrux).

  As part of that, implementing the DH function takes `&mut self` rather than `&mut impl plumbing::ec::EcPrimitives<P256>` -- but Self implements Plumbing, and that is what should be passed to embedded_cal_integration.

* Initially or naively, that wrapper will only be able to provide some EC algorithm if the extendee also implements the P256 (or whichever) curve. This gap can be closed if the extender conditionally forwards EC plumbing either to its extendee or into an implementation of the plumbing based on embedded-cal-iot; embedded-cal-software-demo has an example around that.

WilliamTakeshi:
Didn't had time to review, but just confirming:
I tested here on both boards, every test is working

malishav:
> IIUC we're more preparing this to be ready as a demonstrator than for immediate merging, 

It would be great if we could have this merged this week in order to complete the formal task from MoU and point to this PR.

chrysn:
The merge conflicts stem from this being built on an earlier commit that has been rebased (eg. it has "libcrux-aes: use libcrux-iot version of aesgcm" as dcbc43452077f2ce714c6eb7d399857b89622de0 whereas 2fd30fe49232e15d285a263e1938d64646b85ff3 is in main since #95).

Looking into whether I can just quickly rebase that…

chrysn:
Rebasing was a bit tricky because d6e06928bde7606f72ce60298c00bab37226452e didn't just have content changes but also unrelated formatting changes. Checking that in #120 now.

PR #121, opened 2026-09-16
fix nrf54 probe runner

WilliamTakeshi:
on nrf54l15, everytime I flashed a board for the first time with an image, I got the error:

```console
  Error: An error with the flashing procedure has occurred.

  Caused by:
      Flash content verification failed.
```

to test on hardware, I just ran `cargo test` twice and everything seems to work fine.

but when testing #120, we now build and flash 2 different images, so if one work, the other one doesn't work. So I looked around and seems there is a [workaround](https://github.com/probe-rs/probe-rs/issues/3775): `--disable-double-buffering`


malishav (approved):
just tested this on my setup, where I had the same problem. it seems to fix it!

PR #123, opened 2026-09-17
Move WithRng to dedicated crate

chrysn:
Building on #119, this completes the refactoring of WithRng so that it is usable beyond libcrux, leaving the latter with less clutter.

As a draft it is incomplete; what happened so far was just the moving. What I like to still do is

* [x] Add features so that WithRng also has a new() function that takes the system RNG (via rand::std or getrandom, not sure)
* [x] Uses that to simplfy test construction
* [x] Provides a standalone constructor for both
  * [x] libcrux and
  * [ ] rustcrypto impls that takes the system RNG. (Those then enable reasonably easy use from std systems).

Closes: #119


chrysn:
Can't do that easily with rustcrypto yet because that has some odd internal dependencies (old rand_core version, depending on knowing that it's running on its own RNG even though that's not necessarily realistic).

Cleaning this up for review now.

robinhundt (commented):
Thanks for taking over these changes! Working with stacked PRs from a fork of a repo seems to be a bit of a hassle. I think #119 can be closed in favor of this PR, right?

I like the `WithSysRng` alias/helper method but have a small comment on the naming/docs.

embedded-cal-rand/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+#[cfg(feature = "with_sys")]
+impl<C> WithRng<C, rand::rngs::StdRng> {
+    /// Constructs an extender that uses an [`rand::rngs::StdRng`] seeded from the system.
+    pub fn new_with_sys(cal: C) -> Self {
+        WithRng {
+            cal,
+            rng: rand::make_rng(),
+        }
+    }
+}
```
robinhundt, line 42: The naming and docs are a bit imprecise/conflicting here. The type alias and method is called `with_sys`, but the rng is a `StdRng` and not a `SysRng` as I would initially expect.
The `StdRng` constructed via `rand::make_rng()` is not necessarily seeded directly by the `SysRng`, but may be seeded by a `ThreadRng` if the feature is enabled (see [here](https://docs.rs/rand/latest/src/rand/lib.rs.html#103-113)).

chrysn:
I've set this PR to auto-close 119 once it gets merged.

Not seeing the comments yet: Right, names should be cleaned up / unified a bit.

robinhundt:
My internet is a bit spotty and it seems the comment wasn't submitted. Should be there now.

embedded-cal-rand/src/lib.rs
chrysn: It might be *seeded* by the thread RNG, but that in term is seeded by the sys RNG.

The point here is not so much what is used from the rand crate precisely, but that it is the system RNG from where this comes from; the latest commits reflect that.

chrysn:
Labelling as MVP-critical because docs (#54) can only give sensible guidance if things are reasonably simple here.

embedded-cal-rand/src/lib.rs
robinhundt: Yes, the `thread_rng` is also seeded by the sys RNG, but it has different [security implications/trade-offs][tr] than the sys RNG. In practice, it is probably fine to use `make_rng` (and by that `thread_rng`) here. I just wanted to make sure that this is a conscious decision and not accidental.

The new naming is better.

[tr]: https://docs.rs/rand/latest/rand/rngs/struct.ThreadRng.html#security

PR #124, opened 2026-09-19
rustcrypto: Update and build on underlying WithRng for renamed standalone operation

chrysn:
Depends on: #123

Some steps in #123 couldn't easily be applied because of RNG usage mismatches. Those are fixed in this PR by updating all our dependencies, and then doing #123 was just a matter of setting the right constraints, and eventually renaming the public-facing types for consistency.


chrysn:
While not fully trivial, there was some thumbs-up in chat on merging this without further review, given that it's all just either updating a dependency version or doing something to -rustcrypt that has already been done in -libcrux.

